PR #30, opened 2022-01-27
Refactor helper methods

VijayanB:
### Description
1. Use JSONObject to build Feature
2. Create Parser and move parser methods.
 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/GeospatialParser.java
+++ b/src/main/java/org/opensearch/geospatial/GeospatialParser.java
@@ -15,11 +15,13 @@ import java.util.Map;
 public class GeospatialParser {
 
     /**
-     * Convert Object into {@link Map<String,Object>}
+     * Convert object into Map
+     * @param input Object that is also an instance of Map
+     * @return input object in Map type
      */
     public static Map<String, Object> toStringObjectMap(Object input) {
         if (!(input instanceof Map)) {
-            throw new IllegalArgumentException(input + " is not an instance of Map");
+            throw new IllegalArgumentException(input + " is not an instance of Map, but of type [ " + input.getClass().getName() + " ]");
         }
         Map<Object, Object> inputMap = (Map<Object, Object>) input;
         Map<String, Object> stringObjectMap = new HashMap<>();
--- a/src/main/java/org/opensearch/geospatial/GeospatialParser.java
+++ b/src/main/java/org/opensearch/geospatial/GeospatialParser.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class GeospatialParser {
+
+    public static Map<String, Object> toStringObjectMap(Object input) {
+        if (!(input instanceof Map)) {
+            throw new IllegalArgumentException(input + " is not an instance of Map");
+        }
+        Map<Object, Object> inputMap = (Map<Object, Object>) input;
+        Map<String, Object> stringObjectMap = new HashMap<>();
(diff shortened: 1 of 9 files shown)


src/test/java/org/opensearch/geospatial/GeospatialRestTestCase.java
```diff
@@ -61,9 +58,9 @@ protected static void deletePipeline(String name) throws IOException {
     }
 
     protected static void createIndex(String name, Settings settings, Map<String, String> fieldMap) throws IOException {
-        XContentBuilder xContentBuilder = XContentFactory.jsonBuilder().startObject().startObject(Feature.PROPERTIES_KEY);
```
jmazanec15, line 64: Why hard code these?

VijayanB: Though the value of constant is what i wanted, constant name is not appropriate for the value. I created new constant (duplicate value) with valid name. Do you have any suggestion for how to handle this case? i am not convinced of this method but don't know what is the alternative.

src/main/java/org/opensearch/geospatial/GeospatialParser.java
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class GeospatialParser {
```
martin-gaievski, line 11: What's the purpose of this class, maybe you can put some javadoc description? For now I see only one method that to me is doing more on a conversion/transformation rather than parsing. Are going to add more parser related functionality to this class later?

src/test/java/org/opensearch/geospatial/GeospatialRestTestCase.java
jmazanec15: Ah oh I mixed this up. I thought this was a source file. Im okay with either approach.

src/test/java/org/opensearch/geospatial/GeospatialObjectBuilder.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.opensearch.geospatial;
+
+import java.util.Map;
+
+import org.json.JSONObject;
+import org.opensearch.geospatial.geojson.Feature;
+
+public class GeospatialObjectBuilder {
```
jmazanec15, line 17: Why not use builder pattern for Geospatial object?

VijayanB: Sorry, i was not clear. I updated the description. This is ObjectBuilder class for test. I add this so that both unit test and integration test can use this for input.

src/test/java/org/opensearch/geospatial/processor/FeatureProcessorTests.java
```diff
@@ -21,21 +22,10 @@
...
-        geoJSON.put("properties", properties);
-
-        Map<String, Object> geometry = new HashMap<>();
-        geometry.put(GEOMETRY_TYPE_KEY, GeoShapeType.POINT.shapeName());
-        geometry.put(GEOMETRY_COORDINATES_KEY, "[125.6, 10.1]");
-        geoJSON.put(Feature.GEOMETRY_KEY, geometry);
-
-        return geoJSON;
-
+        return buildGeoJSONFeature(buildGeometry(GeoShapeType.POINT.shapeName(), "[125.6, 10.1]"), buildProperties(properties)).toMap();
```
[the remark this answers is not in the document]
VijayanB: Refactored to generate random point.

martin-gaievski (approved):
Thank you, Lgtm

jmazanec15 (approved):

PR #32, opened 2022-01-28
Create UploadGeoJSONRequest content as an object

VijayanB:
Signed-off-by: Vijayan Balasubramanian <balasvij@amazon.com>

### Description
Create Request content object to deserialize into an object. This will be used by Transport action
to build logic to index request content as document into an index.
 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/GeospatialParser.java
+++ b/src/main/java/org/opensearch/geospatial/GeospatialParser.java
@@ -7,19 +7,24 @@ package org.opensearch.geospatial;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
+
+import org.opensearch.common.bytes.BytesReference;
+import org.opensearch.common.xcontent.XContentHelper;
+import org.opensearch.common.xcontent.XContentType;
 
 /**
  * GeospatialParser provides helper methods to parse/extract/transform input to
  * desired formats.
  */
-public class GeospatialParser {
+public final class GeospatialParser {
 
     /**
      * Convert object into Map
      * @param input Object that is also an instance of Map
      * @return input object in Map type
      */
-    public static Map<String, Object> toStringObjectMap(Object input) {
+    public static Map<String, Object> toStringObjectMap(final Object input) {
         if (!(input instanceof Map)) {
             throw new IllegalArgumentException(input + " is not an instance of Map, but of type [ " + input.getClass().getName() + " ]");
         }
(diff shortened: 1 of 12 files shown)


src/main/java/org/opensearch/geospatial/GeospatialParser.java
```diff
@@ -7,19 +7,24 @@
...
+import org.opensearch.common.bytes.BytesReference;
+import org.opensearch.common.xcontent.XContentHelper;
+import org.opensearch.common.xcontent.XContentType;
 
 /**
  * GeospatialParser provides helper methods to parse/extract/transform input to
  * desired formats.
  */
-public class GeospatialParser {
+public final class GeospatialParser {
```
martin-gaievski, line 20: why we need final here?

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequestContent.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static final ParseField FIELD_INDEX = new ParseField("index", new String[0]);
+    public static final ParseField FIELD_GEOSPATIAL = new ParseField("field", new String[0]);
+    public static final ParseField FIELD_DATA = new ParseField("data", new String[0]);
+    private final String indexName;
+    private final String geospatialFieldName;
+    private final Object data;
+
+    private UploadGeoJSONRequestContent(String indexName, String geospatialFieldName, Object data) {
+        this.indexName = requireNonNull(indexName);
```
martin-gaievski, line 25: I suggest we call requireNonNull with second param and put some explicit message about what param value shouldn't be null

src/test/java/org/opensearch/geospatial/GeospatialObjectBuilder.java
```diff
@@ -21,6 +25,9 @@
```
martin-gaievski, line 25: Is this a utility class for building data for test cases?

VijayanB: Yes, thats correct.

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequestContent.java
VijayanB: Ack

src/main/java/org/opensearch/geospatial/GeospatialParser.java
VijayanB: To make sure no one can override this .

martin-gaievski: This statement is to question :D 
I mean is there a reason why it has been added just now, not with initial change?

VijayanB: i missed it in my previous commit :D . Do you prefer separate commit or PR?

martin-gaievski: nope that's fine, just want to understand the reason, thank you!

src/main/java/org/opensearch/geospatial/GeospatialParser.java
```diff
@@ -30,4 +35,34 @@
...
+    /**
+     * User inputs are usually deserialized into Map. extractValueAsString will help caller to
+     * extract value from the Map and cast it to string with validation.
+     * @param input User input of type Map
+     * @param key property we would like to extract value of
+     * @return null if key doesn't exist, value as String if it exists, throw exception otherwise
+     */
+    public static String extractValueAsString(final Map<String, Object> input, final String key) {
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(input);
```
martin-gaievski, line 48: can we do the same addition with descriptive user message here as well?

martin-gaievski (approved):
Lgtm, thank you

src/main/java/org/opensearch/geospatial/GeospatialParser.java
VijayanB: Ack

jmazanec15 (approved):

PR #33, opened 2022-01-29
Add GeoJSON object of type FeatureCollection

VijayanB:
### Description
Feature Collection represents GeoJSON of type "FeatureCollection".
A FeatureCollection object has a member with the name "features" of type Feature,
which can be empty.
 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/GeospatialParser.java
+++ b/src/main/java/org/opensearch/geospatial/GeospatialParser.java
@@ -5,13 +5,21 @@
 
 package org.opensearch.geospatial;
 
+import static org.opensearch.geospatial.geojson.Feature.TYPE_KEY;
+
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.opensearch.common.bytes.BytesReference;
 import org.opensearch.common.xcontent.XContentHelper;
 import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.geospatial.geojson.Feature;
+import org.opensearch.geospatial.geojson.FeatureCollection;
 
 /**
  * GeospatialParser provides helper methods to parse/extract/transform input to
@@ -64,5 +72,23 @@ public final class GeospatialParser {
     public static Map<String, Object> convertToMap(BytesReference content) {
         Objects.requireNonNull(content);
         return XContentHelper.convertToMap(content, false, XContentType.JSON).v2();
+    }
+
+    /**
+     * getFeatures will return features from given map input. This function abstracts the logic to parse given input and returns
+     * list of Features if exists in Map format.
+     * @param geoJSON given input which may contain GeoJSON Object
+     * @return List of Feature in Map
+     */
+    public static Optional<List<Map<String, Object>>> getFeatures(final Map<String, Object> geoJSON) {
+        final String type = extractValueAsString(geoJSON, TYPE_KEY);
+        Objects.requireNonNull(type, TYPE_KEY + " cannot be null");
+        if (Feature.TYPE.equalsIgnoreCase(type)) {
+            return Optional.ofNullable(Collections.unmodifiableList(Arrays.asList(geoJSON)));
+        }
+        if (FeatureCollection.TYPE.equalsIgnoreCase(type)) {
+            return Optional.ofNullable(Collections.unmodifiableList(FeatureCollection.create(geoJSON).getFeatures()));
+        }
+        return Optional.empty();
     }
 }
--- a/src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
+++ b/src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
@@ -43,9 +43,10 @@ public final class FeatureCollection {
     /**
      * Add Features to this collection
      * @param features List of {@link Feature}
+     * @throws NullPointerException if feature is null
      */
     public void addFeatures(List<Feature> features) {
-        Objects.requireNonNull(features);
+        Objects.requireNonNull(features, "cannot add null to features");
         this.features.addAll(features);
     }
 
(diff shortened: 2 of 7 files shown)


VijayanB:
This will be rebased after #32 is merged.

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
```diff
@@ -0,0 +1,89 @@
...
+    public List<Feature> getFeatures() {
+        return Collections.unmodifiableList(features);
+    }
+
+    /**
+     * Add Features to this collection
+     * @param features List of {@link Feature}
+     */
+    public void addFeatures(List<Feature> features) {
+        Objects.requireNonNull(features);
```
martin-gaievski, line 48: Can we make it follow the pattern from other requireNonNull calls by adding message argument for case when value is actually null?

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * The static method to create an instance of {@link FeatureCollection} from input.
+     *
+     * @param input the object from where {@link FeatureCollection} will be extracted
+     * @return FeatureCollection Instance from given input
+     */
+    public static FeatureCollection create(final Map<String, Object> input) {
+        Object geoJSONType = input.get(TYPE_KEY);
```
martin-gaievski, line 59: Do we need to add check for null for input variable?

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
VijayanB: Definitely. I updated accordingly.

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
VijayanB: Added.

martin-gaievski (approved):
Lgtm, thank you

src/main/java/org/opensearch/geospatial/GeospatialParser.java
```diff
@@ -65,4 +72,22 @@ public static String extractValueAsString(final Map<String, Object> input, final
...
+    public static List<Map<String, Object>> getFeatures(final Map<String, Object> geoJSON) {
+        final String type = extractValueAsString(geoJSON, TYPE_KEY);
+        Objects.requireNonNull(type, TYPE_KEY + " cannot be null");
+        if (Feature.TYPE.equalsIgnoreCase(type)) {
+            return Collections.unmodifiableList(Arrays.asList(geoJSON));
+        }
+        if (FeatureCollection.TYPE.equalsIgnoreCase(type)) {
+            return Collections.unmodifiableList(FeatureCollection.create(geoJSON).getFeatures());
+        }
+        return null;
```
jmazanec15, line 91: Should this be exception?

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
```diff
@@ -0,0 +1,88 @@
...
+import org.opensearch.geospatial.GeospatialParser;
+
+/**
+ * FeatureCollection represents GEOJSON of type FeatureCollection. A FeatureCollection object has a member
+ * with the name "features".  The value of "features" is a List.
+ * Each element of the list is a {@link Feature} object.  It
+ * is possible for this list to be empty.
+ */
+public final class FeatureCollection {
+    public static final String TYPE = "FeatureCollection";
```
jmazanec15, line 24: Typically, OpenSearch uses camel_case. Why not do the same here? Is this a geo convention?

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
```diff
@@ -0,0 +1,88 @@
...
+    private final List<Map<String, Object>> features;
+
+    private FeatureCollection() {
+        this.features = new ArrayList<>();
+    }
+
+    /**
+     * Gets the list of Features in Map format
+     *
+     * @return List of }
```
jmazanec15, line 36: nit: typo?

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
```diff
@@ -0,0 +1,88 @@
...
+ * FeatureCollection represents GEOJSON of type FeatureCollection. A FeatureCollection object has a member
+ * with the name "features".  The value of "features" is a List.
+ * Each element of the list is a {@link Feature} object.  It
+ * is possible for this list to be empty.
+ */
+public final class FeatureCollection {
+    public static final String TYPE = "FeatureCollection";
+    public static final String FEATURES_KEY = "features";
+    public static final String TYPE_KEY = "type";
+    private final List<Map<String, Object>> features;
```
jmazanec15, line 27: Why isnt this a list of Features?

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
VijayanB: Yeah. This value is from GeoJSON RFC.

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
VijayanB: The strategy is , user will pass input as an array of Map. This input will be GeoJSON and it can be either Feature or FeatureCollection. From, API POV, it only needs Feature. So, if user passes FeatureCollection, we only need Map of Features. Hence, there is no need to convert this map to Feature.

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
VijayanB: Ack. Will update.

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
VijayanB: Oops. I had that as Feature and later when i started using this in next PR, i realized i don't need to convert, but forgot to update documentation. Corrected now.

src/main/java/org/opensearch/geospatial/GeospatialParser.java
VijayanB: Added Optional<T> as return type

jmazanec15 (approved):

PR #34, opened 2022-02-04
Include new route to support update index while upload

### LINKED ISSUE #29: Implement UploadGeoJSONRequest
The request should support following input
1. index name
2. Feature Collection
3. geosptial field type (geo_shape or geo_point)
4. Properties

VijayanB:
### Description
1. User might want to append new GeoJSON Features to existing index. Hence, add new route using PUT. 
This route will accept same parameter as POST, except, it ignores, if index already exists. We need different API to support
users from not accidentally insert geoJSON to existing index and intentionally support append geoJSON to existing index. 
2. Add new field called "feature_id" , to identify, most commonly used identifier in GeoJSON as feature id. This will be used as id, while indexing GeoJSON if it exists.
 

### Issues Resolved
#29
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/GeospatialParser.java
+++ b/src/main/java/org/opensearch/geospatial/GeospatialParser.java
@@ -78,7 +78,7 @@ public final class GeospatialParser {
      * getFeatures will return features from given map input. This function abstracts the logic to parse given input and returns
      * list of Features if exists in Map format.
      * @param geoJSON given input which may contain GeoJSON Object
-     * @return List of Feature in Map
+     * @return Optional List of Feature in Map, if Feature exist
      */
     public static Optional<List<Map<String, Object>>> getFeatures(final Map<String, Object> geoJSON) {
         final String type = extractValueAsString(geoJSON, TYPE_KEY);
--- a/src/main/java/org/opensearch/geospatial/GeospatialParser.java
+++ b/src/main/java/org/opensearch/geospatial/GeospatialParser.java
@@ -78,7 +78,8 @@ public final class GeospatialParser {
      * getFeatures will return features from given map input. This function abstracts the logic to parse given input and returns
      * list of Features if exists in Map format.
      * @param geoJSON given input which may contain GeoJSON Object
-     * @return Optional List of Feature in Map, if Feature exist
+     * @return Returns an Optional with the List of Features as map if input is GeoJSON,
+     * else, returns an empty Optional instance.
      */
     public static Optional<List<Map<String, Object>>> getFeatures(final Map<String, Object> geoJSON) {
         final String type = extractValueAsString(geoJSON, TYPE_KEY);
--- a/src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
+++ b/src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
@@ -19,22 +19,36 @@ import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.common.bytes.BytesReference;
 import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
+import org.opensearch.rest.RestRequest;
 
 public class UploadGeoJSONRequest extends ActionRequest {
 
+    private final RestRequest.Method method;
     private BytesReference content;
 
-    public UploadGeoJSONRequest(BytesReference content) {
-        this.content = Objects.requireNonNull(content);
+    public UploadGeoJSONRequest(RestRequest.Method method, BytesReference content) {
+        super();
+        this.method = Objects.requireNonNull(method, "method cannot be null");
+        this.content = Objects.requireNonNull(content, "content cannot be null");
     }
 
     public UploadGeoJSONRequest(StreamInput in) throws IOException {
         super(in);
         this.content = Objects.requireNonNull(in.readBytesReference());
+        this.method = in.readEnum(RestRequest.Method.class);
     }
 
     public BytesReference getContent() {
         return content;
+    }
+
+    /**
+     * getter for Request method type. This will be useful to decide whether new index should
+     * be created for upload request or not.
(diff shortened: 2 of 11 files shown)


VijayanB:
This will be rebased after #33

src/main/java/org/opensearch/geospatial/GeospatialParser.java
```diff
@@ -78,7 +78,7 @@ public static String extractValueAsString(final Map<String, Object> input, final
      * getFeatures will return features from given map input. This function abstracts the logic to parse given input and returns
      * list of Features if exists in Map format.
      * @param geoJSON given input which may contain GeoJSON Object
-     * @return List of Feature in Map
+     * @return Optional List of Feature in Map, if Feature exist
```
martin-gaievski, line 81: I think from current description it isn't clear what we're going to return in case Feature doesn't exist.

src/test/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequestTests.java
```diff
@@ -36,17 +40,22 @@ private String getRequestBody(Map<String, Object> contentMap) {
 
     public void testStreams() throws IOException {
         String requestBody = getRequestBody(null);
-        UploadGeoJSONRequest request = new UploadGeoJSONRequest(new BytesArray(requestBody.getBytes(StandardCharsets.UTF_8)));
+        RestRequest.Method method = PUT;
```
martin-gaievski, line 43: nit: why we use variable to keep request method in this test case, but use constant direct in other test case? I think we the same approach is used in both methods is makes code look more consistent.

VijayanB: This test case checks request input parameters and i am testing getMethod() by asserting down whether method is same as actual or not. In other test case, it was merely an input parameter.

src/main/java/org/opensearch/geospatial/GeospatialParser.java
VijayanB: Sure. Will update.

src/main/java/org/opensearch/geospatial/rest/action/upload/geojson/RestUploadGeoJSONAction.java
```diff
@@ -40,20 +42,53 @@ public String getName() {
...
+     *          "prop0": "value0",
+     *          "prop1": {
+     *             "this": "that"
+     *          }
+     *      }
+     *   }
+     *  ]
      * }
-     *
+     * PUT /_plugins/geospatial/geojson/_upload
```
jmazanec15, line 72: General question: Why do we need the _upload portion for these requests? Why cant we just use `/_plugins/geospatial/geojson`

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequestContent.java
```diff
@@ -21,16 +21,19 @@
...
     private final String fieldType;
+    private final String featureId;
     private final Object data;
 
-    private UploadGeoJSONRequestContent(String indexName, String fieldName, String fieldType, Object data) {
+    private UploadGeoJSONRequestContent(String indexName, String fieldName, String fieldType, String featureId, Object data) {
         this.indexName = indexName;
         this.fieldName = fieldName;
         this.fieldType = fieldType;
+        this.featureId = featureId; // can be null
```
jmazanec15, line 36: In what case would this be null? Also, where will this be used?

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
```diff
@@ -19,24 +19,38 @@
...
+    public UploadGeoJSONRequest(RestRequest.Method method, BytesReference content) {
+        super();
+        this.method = Objects.requireNonNull(method, "method cannot be null");
+        this.content = Objects.requireNonNull(content, "content cannot be null");
     }
 
     public UploadGeoJSONRequest(StreamInput in) throws IOException {
         super(in);
         this.content = Objects.requireNonNull(in.readBytesReference());
+        this.method = in.readEnum(RestRequest.Method.class);
```
jmazanec15, line 38: For consistency, should this be wrapped with requireNonNull?

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequestContent.java
VijayanB: Feature can have field similar to "_id" for Document. In that case, we will use that value for Doc id while uploading

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
VijayanB: Ack

src/main/java/org/opensearch/geospatial/rest/action/upload/geojson/RestUploadGeoJSONAction.java
VijayanB: Ack.

jmazanec15: Im wondering  if there are other actions here that would warrant a post/put just as much as uploading. Can you think of any? 

I guess geojson isnt a resource. It seems more to me like a function. So maybe _upload does make sense. What are your thoughts @VijayanB ?

VijayanB: Let's keep _upload, similar to _bulk where it creates multiple resources based on payload.

jmazanec15 (commented):
Will approve once _upload is added back

jmazanec15 (approved):

PR #58, opened 2022-05-03
Move url path builder to common package

VijayanB:
### Description
Refactor URL path builder to new package to break cyclic dependency.
 
 
### Check List


jmazanec15 (commented):
Was this causing an issue?

VijayanB:
> Was this causing an issue?

It was not an issue so far thats because we don't check for this cyclic dependency. Though it is not harmful, it is not good practice to have cyclic dependency if we can avoid it. Here, the dependency is just building plugin path, hence, i see an opportunity to remove this dependency.

jmazanec15 (approved):

Shivamdhar (approved):

PR #59, opened 2022-05-03
Add support to Writeable interface

VijayanB:
### Description
UploadStats and UploadMetric will be passed between Nodes during StatsRequest. This request will use
StreamInput/StreamOutput to read stats across each other.
 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/stats/upload/UploadStats.java
+++ b/src/main/java/org/opensearch/geospatial/stats/upload/UploadStats.java
@@ -7,7 +7,6 @@ package org.opensearch.geospatial.stats.upload;
 
 import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Objects;
 import java.util.Set;
 
@@ -17,18 +16,6 @@ import org.opensearch.common.metrics.CounterMetric;
  * Contains the total upload stats
  */
 public final class UploadStats {
-
-    public enum FIELDS {
-
-        METRICS,
-        TOTAL,
-        UPLOAD;
-
-        @Override
-        public String toString() {
-            return name().toLowerCase(Locale.getDefault());
-        }
-    }
 
     private static final UploadStats instance = new UploadStats();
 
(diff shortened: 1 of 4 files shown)


src/test/java/org/opensearch/geospatial/stats/upload/UploadStatsTests.java
```diff
@@ -5,12 +5,16 @@
 
 package org.opensearch.geospatial.stats.upload;
 
+import static org.hamcrest.collection.IsIn.in;
```
jmazanec15, line 8: Did you add this on purpose?

src/main/java/org/opensearch/geospatial/stats/upload/UploadStats.java
```diff
@@ -47,6 +38,19 @@ public static UploadStats getInstance() {
         totalAPICount = new CounterMetric();
     }
 
+    /**
+     * Get UploadStats from {@link StreamInput}.
+     * @param input contains {@link UploadStats} in serialized form
+     * @return UploadStats instance
+     * @throws IOException if cannot read {@link UploadStats} from given input
+     */
+    public static UploadStats fromStreamInput(StreamInput input) throws IOException {
```
jmazanec15, line 47: Im wondering if this makes more sense as a constructor of UploadStats. Did you think about this?

src/test/java/org/opensearch/geospatial/stats/upload/UploadStatsTests.java
VijayanB: It is supposed to be cleaned by spotApply :( I removed it now.

src/main/java/org/opensearch/geospatial/stats/upload/UploadStats.java
VijayanB: Right. I thought about it. I chose this over other because of "throws IOException". If it is going to just save the parameter as member input, i would have went with constructor format. What do you think?

jmazanec15: Hmm not sure I understand. Is it bad practice to have an exception thrown from constructor?

jmazanec15 (approved):

PR #70, opened 2022-06-11
Add shape field mapper

VijayanB:
### Description
Added new content type "xy_shape" to index document with
field of type "xy_shape". This is similar to geo_shape. It accepts
value as either GeoJSON format or WKT format.
This has same parameters as geo_shape like ignore_malformed,
ignore_z_value, coerce, orientation.
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldMapper.java
+++ b/src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldMapper.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial.index.mapper.xyshape;
+
+import java.util.List;
+import java.util.Map;
+
+import org.apache.lucene.document.FieldType;
+import org.apache.lucene.index.IndexOptions;
+import org.apache.lucene.index.IndexableField;
+import org.opensearch.common.Explicit;
+import org.opensearch.common.geo.GeometryParser;
+import org.opensearch.geometry.Geometry;
+import org.opensearch.geometry.GeometryVisitor;
+import org.opensearch.index.mapper.AbstractShapeGeometryFieldMapper;
+import org.opensearch.index.mapper.GeoShapeParser;
+import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.index.mapper.ParseContext;
+
+/**
+ *  FieldMapper for indexing {@link org.apache.lucene.document.XYShape}s.
+ */
+public class XYShapeFieldMapper extends AbstractShapeGeometryFieldMapper<Geometry, Geometry> {
+
+    public static final String CONTENT_TYPE = "xy_shape";
+    private static final FieldType FIELD_TYPE = new FieldType();
+    // Similar to geo_shape, this field is indexed by encoding it as triangular mesh
+    // and index each traingle as 7 dimension point in BKD Tree
+    static {
+        FIELD_TYPE.setDimensions(7, 4, Integer.BYTES);
+        FIELD_TYPE.setIndexOptions(IndexOptions.DOCS);
+        FIELD_TYPE.setOmitNorms(true);
+        FIELD_TYPE.freeze();
+    }
+
+    private XYShapeFieldMapper(
+        String simpleName,
+        FieldType fieldType,
+        MappedFieldType mappedFieldType,
+        Explicit ignoreMalformed,
+        Explicit coerce,
+        Explicit ignoreZValue,
+        Explicit orientation,
+        MultiFields multiFields,
+        CopyTo copyTo
+    ) {
+        super(simpleName, fieldType, mappedFieldType, ignoreMalformed, coerce, ignoreZValue, orientation, multiFields, copyTo);
+    }
+
+    @Override
+    public XYShapeFieldType fieldType() {
+        return (XYShapeFieldType) super.fieldType();
+    }
+
+    @Override
+    protected boolean docValuesByDefault() {
+        return false;
+    }
+
+    @Override
+    protected void mergeGeoOptions(AbstractShapeGeometryFieldMapper mergeWith, List conflicts) {
+        // Cartesian plane don't have to support this feature
+    }
+
+    @Override
+    protected String contentType() {
+        return CONTENT_TYPE;
+    }
+
+    @Override
+    protected void addStoredFields(ParseContext context, Geometry geometry) {
+        // No stored fields will be added
+    }
+
+    @Override
+    protected void addDocValuesFields(String name, Geometry geometry, List<IndexableField> fields, ParseContext context) {
+        // doc values are not supported
+    }
+
+    @Override
+    protected void addMultiFields(ParseContext context, Geometry geometry) {
+        // No other fields will be added
+    }
+
+    /**
+     * Builder class to create an instance of {@link XYShapeFieldMapper}
+     */
+    public static class XYShapeFieldMapperBuilder extends AbstractShapeGeometryFieldMapper.Builder<
+        XYShapeFieldMapperBuilder,
+        XYShapeFieldType> {
+
+        public XYShapeFieldMapperBuilder(String fieldName) {
+            super(fieldName, FIELD_TYPE);
+            this.hasDocValues = false;
+        }
+
+        @Override
+        public XYShapeFieldMapper build(BuilderContext context) {
+            return new XYShapeFieldMapper(
+                name,
+                fieldType,
+                buildShapeFieldType(context),
+                ignoreMalformed(context),
+                coerce(context),
+                ignoreZValue(),
+                orientation(),
+                multiFieldsBuilder.build(this, context),
+                copyTo
+            );
+        }
+
+        private XYShapeFieldType buildShapeFieldType(BuilderContext context) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldTypeParser.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package org.opensearch.geospatial.index.mapper.xyshape;
+
+import java.util.Map;
+
+import org.opensearch.index.mapper.AbstractShapeGeometryFieldMapper;
+import org.opensearch.index.mapper.MapperParsingException;
+
+// XYShapeFieldTypeParser to parse and validate mapping parameters
```
navneet1v, line 13: convert this to java doc.

navneet1v (commented):
As part of good practice let's start adding java docs on the classes and public function. This will make sure that we are able to maintain the code for longer period of time.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldMapper.java
```diff
@@ -0,0 +1,141 @@
...
+import org.opensearch.index.mapper.AbstractShapeGeometryFieldMapper;
+import org.opensearch.index.mapper.GeoShapeParser;
+import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.index.mapper.ParseContext;
+
+// FieldMapper for indexing {@link org.apache.lucene.document.XYShape}s.
+public class XYShapeFieldMapper extends AbstractShapeGeometryFieldMapper<Geometry, Geometry> {
+
+    public static final String CONTENT_TYPE = "xy_shape";
+    public static final FieldType FIELD_TYPE = new FieldType();
```
navneet1v, line 27: any reason for making them public?

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldTypeParser.java
navneet1v: and for all other classes too.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldTypeParser.java
```diff
@@ -0,0 +1,26 @@
...
+    @Override
+    protected AbstractShapeGeometryFieldMapper.Builder newBuilder(String name, Map<String, Object> params) {
+        return new XYShapeFieldMapper.XYShapeFieldMapperBuilder(name);
+    }
+
+    @Override
+    protected boolean parseXContentParameters(String name, Map.Entry<String, Object> entry, Map<String, Object> params)
+        throws MapperParsingException {
+        // we don't have to parse for deprecated parameters
+        return false;
```
navneet1v, line 24: please add what are the deprecated parameters here.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeSupportVisitor.java
```diff
@@ -0,0 +1,78 @@
...
+import org.opensearch.geometry.LinearRing;
+import org.opensearch.geometry.MultiLine;
+import org.opensearch.geometry.MultiPoint;
+import org.opensearch.geometry.MultiPolygon;
+import org.opensearch.geometry.Point;
+import org.opensearch.geometry.Polygon;
+import org.opensearch.geometry.Rectangle;
+
+// Visitor to check whether shape is supported for indexing or not
+public final class XYShapeSupportVisitor implements GeometryVisitor<Geometry, RuntimeException> {
```
navneet1v, line 28: what is the purpose of these visitors? For GeoShape it made sense as we were normalizing the lat lon present in the Shape.

Also, if we are only doing null checks cannot we do this with 1 single function?

VijayanB: Super class will call prepare indexing method to determine whether to support particular Geometry or not. Here, we are doing null check for supported geometry and raising unSupportedException for other objects

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldTypeParser.java
VijayanB: This method is used by super class for GeoShape to parse parameters and compare with deprecated param list to identify if user passed any of the deprecated param list or not. This is not applicable for XYShape since we don't have any deprecated parameters

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeIndexableFieldsVisitor.java
```diff
@@ -0,0 +1,116 @@
...
+import org.opensearch.geometry.MultiLine;
+import org.opensearch.geometry.MultiPoint;
+import org.opensearch.geometry.MultiPolygon;
+import org.opensearch.geometry.Point;
+import org.opensearch.geometry.Polygon;
+import org.opensearch.geometry.Rectangle;
+import org.opensearch.geometry.ShapeType;
+import org.opensearch.geospatial.index.common.xyshape.XYShapeConverter;
+
+// Visitor to build only supported Shapes into Lucene indexable fields
```
navneet1v, line 35: convert this to java doc.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldTypeParser.java
VijayanB: Added context. This will result in unnecessary parameter check for deprecated parameters which doesn't exists.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeIndexableFieldsVisitor.java
VijayanB: Ack.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldTypeParser.java
navneet1v: I believe we should not override it. Given the case it was added to support the older versions of GeoShapes we should be good, and we will never have the field in our context.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeSupportVisitor.java
VijayanB: Point p = null;
visitor.visit(p)

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldTypeParser.java
VijayanB: Cool. I removed it.

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldMapper.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    protected void addMultiFields(ParseContext context, Geometry geometry) {
+        // No other fields will be added
+    }
+
+    /**
+     * Builder class to create an instance of {@link XYShapeFieldMapper}
+     */
+    public static class XYShapeFieldMapperBuilder extends AbstractShapeGeometryFieldMapper.Builder<
```
martin-gaievski, line 91: is such line end dictated by spotless?

src/test/java/org/opensearch/geospatial/GeospatialRestTestCase.java
```diff
@@ -160,7 +182,8 @@ protected Map<String, Object> getIndexMapping(String index) throws IOException {
         String responseBody = EntityUtils.toString(response.getEntity());
         Map<String, Object> responseMap = createParser(XContentType.JSON.xContent(), responseBody).map();
         MatcherAssert.assertThat(index + " does not exist", responseMap, Matchers.hasKey(index));
-        return (Map<String, Object>) ((Map<String, Object>) responseMap.get(index)).get(FIELD_MAPPINGS_KEY);
+        final Map<String, Object> mappings = (Map<String, Object>) ((Map<String, Object>) responseMap.get(index)).get(FIELD_MAPPINGS_KEY);
+        return (Map<String, Object>) mappings.get(MAPPING_PROPERTIES_KEY);
```
martin-gaievski, line 186: Why do we change initial format by adding one more nested level?

src/main/java/org/opensearch/geospatial/index/mapper/xyshape/XYShapeFieldMapper.java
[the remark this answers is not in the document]
VijayanB: Correct

src/test/java/org/opensearch/geospatial/GeospatialRestTestCase.java
VijayanB: Created new method to extract properties.

martin-gaievski (approved):
Looks good, thx

navneet1v (approved):

PR #74, opened 2022-06-16
Add shape processor

### LINKED ISSUE #67: New Shape Field for indexing and searching documents that contain cartesian geometries
Introduce new Shape Field to support indexing and searching any arbitrary co-ordinates/geometry (polygon, linestring, etc ..) that are non WGS 84 standard.  This will be similar to geo-shape field except it will represent cartesian plane which is not based on earth-fixed terrestrial reference system.  This new field will be based on Lucene field type [XYShape](https://lucene.apache.org/core/8_3_0/sandbox/org/apache/lucene/document/XYShape.html). 

### Scope
1. This filed will support indexing GeoJSON and WKT formats, similar to geo-shape.
2. Will support following optional parameters
          a. ignore_malformed ( default: false )
          b. ignore_z_value ( default: true )
          c. coerce  ( default: false )
          d. orientation ( default : RIGHT/CCW/COUNTERCLOCKWISE)

VijayanB:
### Description

- Create Geometry Visitor pattern to create Lucene XYGeometry objects based on OpenSearch Geometry. This geometry will be either from user defined shape or doc id referrenced from index which contains xy_shape field.
- Processor to accept user input as Geometry and convert it into Lucene Query based on XYShapeVisitor

 
### Issues Resolved
#67 
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
+++ b/src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial.index.query.xyshape;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Objects;
+
+import org.apache.lucene.document.ShapeField;
+import org.apache.lucene.document.XYShape;
+import org.apache.lucene.geo.XYGeometry;
+import org.apache.lucene.search.MatchNoDocsQuery;
+import org.apache.lucene.search.Query;
+import org.opensearch.common.geo.ShapeRelation;
+import org.opensearch.geometry.Geometry;
+import org.opensearch.geometry.GeometryVisitor;
+import org.opensearch.geospatial.index.mapper.xyshape.XYShapeFieldMapper;
+import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.index.query.QueryShardContext;
+import org.opensearch.index.query.QueryShardException;
+
+/**
+ * Query Processor to convert given Geometry into Lucene query
+ */
+public class XYShapeQueryProcessor {
+
+    /**
+     * Creates a {@link Query} that matches all indexed shapes to the provided {@link Geometry}  based on {@link ShapeRelation}
+     * @param geometry OpenSearch {@link Geometry} as an input
+     * @param fieldName Lucene field of type {@link XYShape}
+     * @param relation Relation to be used to get all shapes from given Geometry
+     * @param context QueryShardContext instance
+     * @return {@link Query} instance from XYShape.newGeometryQuery
+     */
+    public Query shapeQuery(
+        Geometry geometry,
+        String fieldName,
+        ShapeRelation relation,
+        GeometryVisitor<List<XYGeometry>, RuntimeException> visitor,
+        QueryShardContext context
+    ) {
+        // if no input is parsed from input, return no documents should be matched from index,
+        // this is similar to geo_shape field type
+        if (geometry == null || geometry.isEmpty()) {
+            return new MatchNoDocsQuery();
+        }
+
+        Objects.requireNonNull(fieldName, "Field name cannot be null");
+        Objects.requireNonNull(context, "QueryShardContext cannot be null");
+        validateIsXYShapeFieldType(fieldName, context);
+
+        Objects.requireNonNull(relation, "ShapeRelation cannot be null");
+        return getQueryFromGeometry(geometry, fieldName, relation.getLuceneRelation(), visitor);
+    }
+
+    private Query getQueryFromGeometry(
+        Geometry geometry,
+        String fieldName,
+        ShapeField.QueryRelation queryRelation,
+        GeometryVisitor<List<XYGeometry>, RuntimeException> visitor
+    ) {
+        final List<XYGeometry> collections = geometry.visit(visitor);
+        // if no valid geometries are found from input shape, return No documents matched query as fallback
+        if (collections == null || collections.isEmpty()) {
+            return new MatchNoDocsQuery();
+        }
+        return XYShape.newGeometryQuery(fieldName, queryRelation, collections.toArray(new XYGeometry[0]));
+    }
+
+    private void validateIsXYShapeFieldType(String fieldName, QueryShardContext context) {
+        MappedFieldType fieldType = context.fieldMapper(fieldName);
+        if (fieldType instanceof XYShapeFieldMapper.XYShapeFieldType) {
+            return;
+        }
+        throw new QueryShardException(
+            context,
+            String.format(
+                Locale.getDefault(),
+                "Expected %s field type for Field [ %s ] but found %s",
+                XYShapeFieldMapper.CONTENT_TYPE,
+                fieldName,
+                fieldType.typeName()
+            )
+        );
+    }
+}
--- a/src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
+++ b/src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
@@ -14,6 +14,7 @@ import static org.opensearch.geospatial.index.common.xyshape.XYShapeConverter.to
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Objects;
 
 import org.apache.lucene.geo.XYGeometry;
@@ -29,12 +30,11 @@ import org.opensearch.geometry.MultiPolygon;
 import org.opensearch.geometry.Point;
 import org.opensearch.geometry.Polygon;
 import org.opensearch.geometry.Rectangle;
-import org.opensearch.geometry.ShapeType;
 import org.opensearch.index.query.QueryShardContext;
 import org.opensearch.index.query.QueryShardException;
 
 /**
- * Geomtery Visitor to create a query to find all cartesian XYShapes
+ * Geometry Visitor to create a query to find all cartesian XYShapes
  * that comply ShapeRelation with all other XYShapes objects
  */
 public class XYShapeQueryVisitor implements GeometryVisitor<List<XYGeometry>, RuntimeException> {
@@ -49,7 +49,7 @@ public class XYShapeQueryVisitor implements GeometryVisitor<List<XYGeometry>, Ru
 
     @Override
     public List<XYGeometry> visit(Circle circle) throws RuntimeException {
-        Objects.requireNonNull(circle, ShapeType.CIRCLE + " cannot be null");
+        Objects.requireNonNull(circle, "Circle cannot be null");
         // XYShape don't support indexing Circle, but, can perform query to identify list of shapes
         // that interact with a provided circle
         return List.of(toXYCircle(circle));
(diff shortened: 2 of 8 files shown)


src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
```diff
@@ -0,0 +1,81 @@
...
+        ShapeRelation relation,
+        GeometryVisitor<List<XYGeometry>, RuntimeException> visitor,
+        QueryShardContext context
+    ) {
+        if (geometry == null || geometry.isEmpty()) {
+            return new MatchNoDocsQuery();
+        }
+        Objects.requireNonNull(geometry, "Field name cannot be null");
+        Objects.requireNonNull(geometry, "ShapeRelation cannot be null");
+        Objects.requireNonNull(geometry, "QueryShardContext cannot be null");
```
vamshin, line 50: looks like these validations should be fieldName, relation, context

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private Query getQueryFromGeometry(
+        Geometry geometry,
+        String fieldName,
+        ShapeField.QueryRelation queryRelation,
+        GeometryVisitor<List<XYGeometry>, RuntimeException> visitor
+    ) {
+        final List<XYGeometry> collections = geometry.visit(visitor);
+        return XYShape.newGeometryQuery(fieldName, queryRelation, collections.toArray(new XYGeometry[0]));
```
vamshin, line 62: Should we check for collections isEmpty?

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Query Processor to convert given Geometry into Lucene query
+ */
+public class XYShapeQueryProcessor {
+
+    /**
+     * Creates a {@link Query} that matches all indexed shapes to the provided {@link Geometry}  based on {@link ShapeRelation}
+     * @param geometry OpenSearch {@link Geometry} as an input
+     * @param fieldName Lucene field of type {@link XYShape}
+     * @param relation Relation to be used get all shapes from giben Geometry
```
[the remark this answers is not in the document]
VijayanB: Ack

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
VijayanB: Good catch. Updated the condition and added unit test to catch this.

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
VijayanB: Good point. Added condition and included unit tests

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+        // validate field type first
+        Objects.requireNonNull(fieldName, "Field name cannot be null");
+        Objects.requireNonNull(context, "QueryShardContext cannot be null");
+        validateIsXYShapeFieldType(fieldName, context);
+
+        // if no input is parsed from input, return no documents should be matched from index,
+        // this is similar to geo_shape field type
+        if (geometry == null || geometry.isEmpty()) {
+            return new MatchNoDocsQuery();
+        }
```
vamshin, line 54: This can be moved to beginning to avoid other validations if geometry is null

vamshin (approved):
LGTM! Thanks

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
VijayanB: Good point. Currently we validate field name and field type first before checking geometry. If geometry is null, we don’t throw exception, instead, will not return any documents from index. Hence I prioritized validation for exception over geometry. WDYT?

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+import org.opensearch.geometry.GeometryVisitor;
+import org.opensearch.geospatial.index.mapper.xyshape.XYShapeFieldMapper;
+import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.index.query.QueryShardContext;
+import org.opensearch.index.query.QueryShardException;
+
+/**
+ * Query Processor to convert given Geometry into Lucene query
+ */
+public class XYShapeQueryProcessor {
```
navneet1v, line 28: does this processor needs to extends/implement some interface? How this processor and which function of this processor is getting called to create the query?

VijayanB: Yes. It will be called by field type. That will be my next PR

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
vamshin: Not a big concern. Just that these validations will be unnecessary if ```geometry``` is null. So if possible better to move to beginning.

VijayanB: Moved this to beginning of the method.

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private Query getQueryFromGeometry(
+        Geometry geometry,
+        String fieldName,
+        ShapeField.QueryRelation queryRelation,
+        GeometryVisitor<List<XYGeometry>, RuntimeException> visitor
+    ) {
+        final List<XYGeometry> collections = geometry.visit(visitor);
+        // if no valid geometries are found from input shpae, return No documents matched query as fallback
```
martin-gaievski, line 66: nit: typo, "shpae" should be "shape"

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private void validateIsXYShapeFieldType(String fieldName, QueryShardContext context) {
+        MappedFieldType fieldType = context.fieldMapper(fieldName);
+        if (fieldType instanceof XYShapeFieldMapper.XYShapeFieldType) {
+            return;
+        }
+        throw new QueryShardException(
+            context,
+            String.format(
+                Locale.getDefault(),
```
martin-gaievski, line 81: I see in core OpenSearch we use ```Locale.ROOT```, not sure what is the best practice here.

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
```diff
@@ -0,0 +1,121 @@
...
+import org.opensearch.geometry.MultiPolygon;
+import org.opensearch.geometry.Point;
+import org.opensearch.geometry.Polygon;
+import org.opensearch.geometry.Rectangle;
+import org.opensearch.geometry.ShapeType;
+import org.opensearch.index.query.QueryShardContext;
+import org.opensearch.index.query.QueryShardException;
+
+/**
+ * Geomtery Visitor to create a query to find all cartesian XYShapes
```
martin-gaievski, line 37: nit: typo "Geomtery" should be "Geometry"

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Override
+    public List<XYGeometry> visit(Line line) throws RuntimeException {
+        Objects.requireNonNull(line, "Line cannot be null");
+        return List.of(toXYLine(line));
+    }
+
+    @Override
+    public List<XYGeometry> visit(LinearRing linearRing) throws RuntimeException {
+        throw new QueryShardException(this.context, "Field [" + this.name + "] found an unsupported shape LinearRing");
```
martin-gaievski, line 72: String.format ?

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
```diff
@@ -0,0 +1,121 @@
...
+    public List<XYGeometry> visit(Circle circle) throws RuntimeException {
+        Objects.requireNonNull(circle, ShapeType.CIRCLE + " cannot be null");
+        // XYShape don't support indexing Circle, but, can perform query to identify list of shapes
+        // that interact with a provided circle
+        return List.of(toXYCircle(circle));
+    }
+
+    @Override
+    public List<XYGeometry> visit(GeometryCollection<?> geometryCollection) throws RuntimeException {
+        Objects.requireNonNull(geometryCollection, " Geometry collection cannot be null");
```
martin-gaievski, line 60: nit: Extra space in error message

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public List<XYGeometry> visit(LinearRing linearRing) throws RuntimeException {
+        throw new QueryShardException(this.context, "Field [" + this.name + "] found an unsupported shape LinearRing");
+    }
+
+    @Override
+    public List<XYGeometry> visit(MultiLine multiLine) throws RuntimeException {
+        Objects.requireNonNull(multiLine, "Multi line cannot be null");
```
martin-gaievski, line 77: nit: should we change "line" to "Line" to have it consistent with other methods

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
VijayanB: Ack

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
VijayanB: Ack

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
VijayanB: Ack

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
VijayanB: Ack

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryVisitor.java
VijayanB: Ack

src/main/java/org/opensearch/geospatial/index/query/xyshape/XYShapeQueryProcessor.java
VijayanB: Will create an issue to investigate and will create new PR if we have to replace to Locale.ROOT. WDYT?

VijayanB: Created issue here: https://github.com/opensearch-project/geospatial/issues/77

martin-gaievski (approved):
Looks good, thx

PR #79, opened 2022-06-20
Add parameter to randomly include z coordinates to geometry

VijayanB:
### Description
Include whether to add z coords into Geometry or not
as param to random function, except Polygon.
Generating true random polygon is very complicated since 3 points at least
to be non-collinear. Will add support to be truly random
during later commit.

Why?
In ShapeQueryBuilderTests (unit test), we will not be processing z coordinates ( xy_shape is 2d object). Hence, needed an
interface to only create Geometry without z coordinates to compare shape output with query input.
These tests are added from super class AbstractQueryTestCase, hence, we cannot modify it. 
 
 
### Check List


vamshin (approved):
LGTM! Thanks

martin-gaievski (approved):
Looks good, thx

PR #88, opened 2022-07-05
Use lombok annotations

VijayanB:
### Description
* Update compile task to recognize lombok
* Use lombok wherever applicable
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
+++ b/src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
@@ -14,6 +14,10 @@ package org.opensearch.geospatial.action.upload.geojson;
 import java.io.IOException;
 import java.util.Objects;
 
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NonNull;
+
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.common.bytes.BytesReference;
@@ -21,34 +25,23 @@ import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
 import org.opensearch.rest.RestRequest;
 
+@AllArgsConstructor
+@Getter
 public class UploadGeoJSONRequest extends ActionRequest {
 
+    /**
+     * Request method type. This will be useful to decide whether new index should
+     * be created for upload request or not.
+     */
+    @NonNull
     private final RestRequest.Method method;
+    @NonNull
     private final BytesReference content;
-
-    public UploadGeoJSONRequest(RestRequest.Method method, BytesReference content) {
-        super();
-        this.method = Objects.requireNonNull(method, "method cannot be null");
-        this.content = Objects.requireNonNull(content, "content cannot be null");
-    }
 
     public UploadGeoJSONRequest(StreamInput in) throws IOException {
         super(in);
         this.content = Objects.requireNonNull(in.readBytesReference(), "data is missing");
         this.method = Objects.requireNonNull(in.readEnum(RestRequest.Method.class), "RestRequest Method is missing");
-    }
-
-    public BytesReference getContent() {
-        return content;
-    }
-
-    /**
-     * getter for Request method type. This will be useful to decide whether new index should
-     * be created for upload request or not.
-     * @return RestRequest.Method either PUT or POST
-     */
-    public RestRequest.Method getMethod() {
-        return method;
     }
 
     @Override
--- a/src/main/java/org/opensearch/geospatial/geojson/Feature.java
+++ b/src/main/java/org/opensearch/geospatial/geojson/Feature.java
@@ -8,9 +8,12 @@ package org.opensearch.geospatial.geojson;
 import java.util.HashMap;
 import java.util.Map;
 
+import lombok.Getter;
+
 /**
  * Feature object represents GEOJSON of type Feature.
  */
+@Getter
 public class Feature {
     public static final String TYPE = "Feature";
     public static final String GEOMETRY_KEY = "geometry";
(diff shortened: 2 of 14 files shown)


src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
```diff
@@ -14,43 +14,37 @@
...
+import lombok.NonNull;
+
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.common.bytes.BytesReference;
 import org.opensearch.common.io.stream.StreamInput;
 import org.opensearch.common.io.stream.StreamOutput;
 import org.opensearch.rest.RestRequest;
 
+@AllArgsConstructor()
```
navneet1v, line 28: () not required.

Also you can move the @Getter on the top of the class. If all attributes of the class requires @Getter

src/main/java/org/opensearch/geospatial/index/common/xyshape/XYShapeConverter.java
```diff
@@ -102,7 +97,7 @@ public static XYCircle toXYCircle(Circle circle) {
      * @param point of type {@link Point}
      * @return {@link XYPoint} instance
      */
-    public static XYPoint toXYPoint(Point point) {
+    public static XYPoint toXYPoint(@NonNull Point point) {
```
navneet1v, line 100: for all @NonNull annotation sending the NullPointerException is bad. I would recommend we should throw IllegalArgumentException. For lombok you can configure this via lombok.config file.

```
lombok.nonNull.exceptionType = IllegalArgumentException
```

VijayanB: Here the value is null, hence it is thrown NPE. Why do you think it is bad? IMO, it will be very intuitive for null object to throw NPE . User will easily know what went wrong if the exception and reason is very clear.

navneet1v: The idea of sending a NPE means the code was trying to accesses data from a Null object, but here how @NonNull annotation is interpreted is like this:
```
if(point == null) {
   throw NullPointerException("parameter is null");
}
```
which doesn't seem fair as we validating the object to be non null. But as the object was null so it make sense to have IllegalArgumentException(any exception which make sense here).

This is little subjective, but as per developer mind set seeing a NPE is bad and people look for cases where object was null and code is trying to access a field from that object which is not this use case.

VijayanB: I would think it is fair, because we are interested in getX()/getY() and we are validating for that, since we want to fail sooner than later. IllegalArgumentException emphasis more on object value than object itself. if input is found null, user can immediately jump to place where it is created or instantiated to easily see what went wrong.

src/main/java/org/opensearch/geospatial/geojson/Feature.java
```diff
@@ -8,9 +8,12 @@
 import java.util.HashMap;
 import java.util.Map;
 
+import lombok.Getter;
+
 /**
  * Feature object represents GEOJSON of type Feature.
  */
+@Getter
```
martin-gaievski, line 16: you can also use @Value here, it has few perks on top of Getter, https://projectlombok.org/features/Value

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
```diff
@@ -14,43 +14,36 @@
...
+     * Request method type. This will be useful to decide whether new index should
+     * be created for upload request or not.
+     */
+    @NonNull
     private final RestRequest.Method method;
+    @NonNull
     private final BytesReference content;
 
-    public UploadGeoJSONRequest(RestRequest.Method method, BytesReference content) {
-        super();
```
navneet1v, line 30: did we check that super will be called by using the AllArgsConstructor?

src/main/java/org/opensearch/geospatial/geojson/FeatureCollection.java
```diff
@@ -17,32 +19,26 @@
  * with the name "features".  The value of "features" is a List.
  * and it is possible for this list to be empty, if {@link FeatureCollection} has no Features.
  */
+@Getter
```
martin-gaievski, line 22: same as above, check if Value annotation is more suitable here

src/main/java/org/opensearch/geospatial/index/common/xyshape/XYShapeConverter.java
navneet1v: "user can immediately jump to place where it is created or instantiated to easily see what went wrong"

This thing will be difficult as the stacktrace will come with the line which user can see only in the delombok code not in the normal code.

Nevertheless, this is quite a subjective topic. Please see if we still feel it should be a NPE instead of IllegalArgumentException

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
martin-gaievski: +1, I think it's not possible for constructor related annotations - https://stackoverflow.com/a/29771875

VijayanB: Yeah. it is not possible. However If a constructor does not explicitly invoke a superclass constructor, the Java compiler automatically inserts a call to the no-argument constructor of the superclass.

navneet1v: Can you please check delombok source here and confirm this?

navneet1v (approved):

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequest.java
VijayanB: It works. I double checked with debugger too.

PR #101, opened 2022-07-28
Include feature branch in workflow to trigger CI

naveentatikonda:
Signed-off-by: Naveen Tatikonda <navtat@amazon.com>

### Description
Cherry-pick e62c0ac22e8f1baa64a0b04d51923b1a41d68b27 from this PR https://github.com/opensearch-project/geospatial/pull/102
 
### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - "*"
+      - "feature/**"
   pull_request:
     branches:
       - "*"
+      - "feature/**"
 
 jobs:
   Build:


.github/workflows/CI.yml
```diff
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - "*"
+      - "feature/**"
   pull_request:
     branches:
       - "*"
+      - "feature/**"
```
navneet1v, line 10: will the CI workflow will run for my forked repos too?

VijayanB: In your forked repo, mostly you will be creating branch without prefix "feature/xyz" and it works by default.

VijayanB:
This should be merged to main as well.

.github/workflows/CI.yml
navneet1v: if it works by default then why we are making these changes. This is confusing.

naveentatikonda:
> This should be merged to main i believe.

I'm creating my feature branches from `feature/xypoint` branch and merging into it. So, I think we should also merge this PR to `feature/xypoint`. What do you think ?

VijayanB:
> > This should be merged to main i believe.
> 
> I'm creating my feature branches from `feature/xypoint` branch and merging into it. So, I think we should also merge this PR to `feature/xypoint`. What do you think ?

I am not 100% sure but having it on main alone is enough I guess.

.github/workflows/CI.yml
naveentatikonda: If my understanding is correct, the branches which we are mentioning in the workflow are the target branches to which we we are creating PRs.
Reference - https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore

VijayanB: The * wildcard matches any character, but does not match slash (/).

naveentatikonda:
> > > This should be merged to main i believe.
> > 
> > 
> > I'm creating my feature branches from `feature/xypoint` branch and merging into it. So, I think we should also merge this PR to `feature/xypoint`. What do you think ?
> 
> I am not 100% sure but having it on main alone is enough I guess.

OK, then I will create a PR to merge to main branch and if it doesn't work then we will merge this PR.

.github/workflows/CI.yml
VijayanB: > @VijayanB if it works by default then why we are making these changes. This is confusing.
> 
> `*` here should taken care of running the workflows for all the branches, but somehow that is not working.
> 
> BTW I am not a big fan of CI/CD workflows running on the forked repos.

There are two events here, push and pull request. CI runs on your forked repo because you will be creating a new branch in order to create a PR. Whereas you will see CI runs as part of PR due to pull request event on main repository.

navneet1v (approved):

VijayanB (approved):

PR #165, opened 2022-10-12
Introduce H3 min resolution as constant

VijayanB:
### Description
H3 version 1 has 16 resolutions, numbered 0 through 15.
Introduced a constant to represent min value, similar
to max value.
 
### Issues Resolved
[List any issues this PR will resolve]
 
### Check List

### DIFF
--- a/libs/h3/src/main/java/org/opensearch/geospatial/h3/H3.java
+++ b/libs/h3/src/main/java/org/opensearch/geospatial/h3/H3.java
@@ -35,6 +35,7 @@ import static java.lang.Math.toRadians;
  */
 public final class H3 {
 
+    public static int MIN_H3_RES = Constants.MIN_H3_RES;
     public static int MAX_H3_RES = Constants.MAX_H3_RES;
 
     /**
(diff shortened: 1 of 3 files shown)


navneet1v (approved):

libs/h3/src/main/java/org/opensearch/geospatial/h3/H3.java
```diff
@@ -35,6 +35,7 @@
  */
 public final class H3 {
 
+    public static int MIN_H3_RES = Constants.MIN_H3_RES;
```
heemin32, line 38: Why do we need this variable?

VijayanB: For consistency, we added this to H3.java similar to MAX_H3_RES.

heemin32: Can we remove MAX_H3_RES instead?

VijayanB: H3 class defines the public API of the H3 library. We imported this PR. I will create an issue to refactor/clean up, for now will keep it as it is. Will that work?

VijayanB: https://github.com/opensearch-project/geospatial/issues/170

heemin32: Fair enough.

heemin32 (approved):

PR #218, opened 2023-02-09
Add limit to geojson upload API

VijayanB:
### Description
Upload API will limit number of features to be uploaded to 10_000. Since custom vector map will allow 10K documents to
be fetched, we will also add same constraint while upload as well for consistency .

Signed-off-by: Vijayan Balasubramanian <balasvij@amazon.com> 
### Issues Resolved
https://github.com/opensearch-project/dashboards-maps/issues/220
 
### Check List


naveentatikonda (approved):

src/main/java/org/opensearch/geospatial/action/upload/geojson/UploadGeoJSONRequestContent.java
```diff
@@ -61,9 +64,28 @@ public static UploadGeoJSONRequestContent create(Map<String, Object> input) {
                 geoJSONData + " is not an instance of List, but of type [ " + geoJSONData.getClass().getName() + " ]"
             );
         }
+        final long featureCount = getFeatureCount(geoJSONData);
+        if (featureCount > MAX_SUPPORTED_GEOJSON_FEATURE_COUNT) {
```
naveentatikonda, line 68: Do you want to add a comment here that why we are setting the max limit to 10_000.

naveentatikonda, line 68: Also, can we break this down into a separate method like `validateFeatureCount(geoJSONData) { ....}` ?

VijayanB: i added comment close to constant

VijayanB: created new method

martin-gaievski (approved):

PR #265, opened 2023-04-26
Update gradle version to 7.6

VijayanB:
### Description
Update gradle version to 7.6
 
### Check List


naveentatikonda:
Is this the reason why 2.6 integ tests are failing for geospatial? If yes, then we might also need to backport this to 2.6 branch

naveentatikonda (approved):

VijayanB:
> Is this the reason why 2.6 integ tests are failing for geospatial? If yes, then we might also need to backport this to 2.6 branch

I don't think so, this is only problem with main.

junqiu-lei (approved):

PR #271, opened 2023-04-27
Added unit tests with some refactoring of codes

heemin32:
### Description
Added unit tests with some refactoring of codes

* Add Unit tests
* Set cache true for search query
* Remove in memory cache implementation (Two way door decision)
  * Relying on search cache without custom cache
* Renamed datasource state from FAILED to CREATE_FAILED
### Issues Resolved
N/A
 
### Check List


src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
```diff
@@ -206,14 +191,46 @@ public void onFailure(final Exception e) {
...
     }
 
+    @VisibleForTesting
+    protected void handleSingleIp(
+        final String ip,
+        final Map<String, Object> stringObjectMap,
+        final IngestDocument ingestDocument,
+        final BiConsumer<IngestDocument, Exception> handler
+    ) {
+        cache.put(ip, datasourceName, stringObjectMap);
```
junqiu-lei, line 201: Can we change name `stringObjectMap` to a more descriptive name?

src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
```diff
@@ -206,14 +191,46 @@ public void onFailure(final Exception e) {
...
+        if (!stringObjectMap.isEmpty()) {
+            ingestDocument.setFieldValue(targetField, filteredGeoData(stringObjectMap, ip));
+        }
+        handler.accept(ingestDocument, null);
+    }
+
+    private Map<String, Object> filteredGeoData(final Map<String, Object> geoData, final String ip) {
+        Map<String, Object> filteredGeoData;
+        if (properties == null) {
+            filteredGeoData = geoData;
```
junqiu-lei, line 211: we can add early return here

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -322,19 +326,31 @@ public boolean isExpired() {
...
+    }
+
+    public boolean isCompatible(final List<String> fields) {
+        Set<String> fieldsSet = new HashSet<>(fields);
+        for (String field : database.fields) {
+            if (!fieldsSet.contains(field)) {
+                return false;
+            }
+        }
+        return true;
```
junqiu-lei, line 344:
```suggestion
/**
 * Checks if the database fields are compatible with the given set of input fields.
 *
 * @param inputFields The set of input fields to check for compatibility.
 * @return Returns true if the database fields are compatible with the given input fields, false otherwise.
 */
public boolean isCompatible(final List<String> inputFields) {
    Set<String> fieldsSet = new HashSet<>(inputFields);
    return areFieldsCompatible(fieldsSet, database.fields); 
}

/**
 * Checks if a set of fields is compatible with another set of fields.
 *
 * @param inputFields The set of input fields to check for compatibility.
 * @param databaseFields The set of database fields to check for compatibility.
 * @return Returns true if the database fields are compatible with the given input fields, false otherwise.
 */
private boolean areFieldsCompatible(Set<String> inputFields, Set<String> databaseFields) {
    return inputFields.containsAll(databaseFields);
}
```

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
```diff
@@ -118,161 +105,42 @@ public void runJob(final ScheduledJobParameter jobParameter, final JobExecutionC
...
-     * @param manifest
-     * @return
-     */
-    private boolean skipUpdate(final Datasource parameter, final DatasourceManifest manifest) {
-        if (manifest.getMd5Hash().equals(parameter.getDatabase().getMd5Hash())) {
-            return true;
+        try {
+            datasourceUpdateService.deleteUnusedIndices(datasource);
+            datasourceUpdateService.updateGeoIpData(datasource);
+            datasourceUpdateService.deleteUnusedIndices(datasource);
```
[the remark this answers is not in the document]
heemin32: The #137 is trying to delete indices which were failed to be deleted from the previous attempt before creating another index.

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
heemin32: `database.fields` is a list not set. Let me add a comment though.

src/main/plugin-metadata/plugin-security.policy
```diff
@@ -9,4 +9,7 @@
 grant {
   // needed by Ip2Geo datasource to get GeoIP database
   permission java.net.SocketPermission "*", "connect,resolve";
+
+  // needed for unit test
+  permission java.io.FilePermission "<<ALL FILES>>", "read";
```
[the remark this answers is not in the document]
heemin32: Let me see if I can scope it down to a certain folder.

src/main/java/org/opensearch/geospatial/annotation/VisibleForTesting.java
```diff
@@ -0,0 +1,12 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.geospatial.annotation;
+
+public @interface VisibleForTesting {
```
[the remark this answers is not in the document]
heemin32: We don't have a dependency on google common. I don't want to add it for this single annotation for now.

heemin32:
Discussed offline with @jmazanec15. He made a good point that I should not register setting update in processor constructor which does not deregister itself when it is deleted. Will resolve the issue.

src/main/plugin-metadata/plugin-security.policy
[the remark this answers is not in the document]
heemin32: It was needed for integration test. Let me remove integration test and this line of code all together.

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataHelper.java
```diff
@@ -61,21 +63,24 @@ public class GeoIpDataHelper {
     private static final String DATA_FIELD_NAME = "_data";
     private static final Tuple<String, Integer> INDEX_SETTING_NUM_OF_SHARDS = new Tuple<>("index.number_of_shards", 1);
     private static final Tuple<String, String> INDEX_SETTING_AUTO_EXPAND_REPLICAS = new Tuple<>("index.auto_expand_replicas", "0-all");
+    private final ClusterService clusterService;
+    private final Client client;
+    private TimeValue timeout;
+
+    public GeoIpDataHelper(final ClusterService clusterService, final Client client) {
```
VijayanB, line 70: nit: Can we rename to GeoIpDataFacade ? Helper classes are usually static.

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceTransportAction.java
```diff
@@ -91,99 +100,106 @@ public PutDatasourceTransportAction(
...
+        datasource.getIndices().add(indexName);
+        datasourceHelper.updateDatasource(datasource);
+        geoIpDataHelper.createIndexIfNotExists(indexName);
         return indexName;
     }
 
-    private String[] putIp2GeoData(final String indexName, final DatasourceManifest manifest) throws IOException {
-        String[] fields;
-        try (CSVParser reader = GeoIpDataHelper.getDatabaseReader(manifest)) {
+    private List<String> putGeoIpData(final String indexName, final DatasourceManifest manifest) throws IOException {
```
VijayanB, line 196: put method returning header seems odd. Shall we split it into two method? put and getHeader for better readability?

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceTransportAction.java
```diff
@@ -91,99 +100,106 @@ public PutDatasourceTransportAction(
...
+        String[] header;
+        try (CSVParser reader = geoIpDataHelper.getDatabaseReader(manifest)) {
             Iterator<CSVRecord> iter = reader.iterator();
-            fields = iter.next().values();
-            GeoIpDataHelper.putGeoData(client, indexName, fields, iter, indexingBulkSize, timeout);
+            header = iter.next().values();
```
VijayanB, line 200: shouldn't we check hasNext() before calling next?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -322,19 +331,43 @@ public boolean isExpired() {
...
+
+    /**
+     * Checks if the database fields are compatible with the given set of fields.
+     *
+     * @param fields The set of input fields to check for compatibility.
+     * @return true if the database fields are compatible with the given input fields, false otherwise.
+     */
+    public boolean isCompatible(final List<String> fields) {
+        Set<String> fieldsSet = new HashSet<>(fields);
+        for (String field : database.fields) {
```
VijayanB, line 356: you can also use size to decide if they are not compatible before checking all fields.

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
```diff
@@ -118,161 +105,42 @@ public void runJob(final ScheduledJobParameter jobParameter, final JobExecutionC
...
-
-        if (skipUpdate(jobParameter, manifest)) {
-            log.info("Skipping GeoIP database update. Update is not required for {}", jobParameter.getId());
-            jobParameter.getUpdateStats().setLastSkippedAt(Instant.now());
-            DatasourceHelper.updateDatasource(client, jobParameter, timeout);
+    @VisibleForTesting
+    protected void updateDatasource(final ScheduledJobParameter jobParameter) throws IOException {
+        Datasource datasource = datasourceHelper.getDatasource(jobParameter.getName());
+        if (datasource == null) {
+            log.info("Datasource[{}] is already deleted", jobParameter.getName());
```
VijayanB, line 132: nit: shall we use term doesn't exists instead of deleted? The first doesn't assume it was previously created

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
```diff
@@ -0,0 +1,181 @@
...
+        String indexName = jobParameter.indexNameFor(manifest);
+        jobParameter.getIndices().add(indexName);
+        datasourceHelper.updateDatasource(jobParameter);
+        geoIpDataHelper.createIndexIfNotExists(indexName);
+        String[] fields;
+        List<String> fieldsToStore;
+        try (CSVParser reader = geoIpDataHelper.getDatabaseReader(manifest)) {
+            Iterator<CSVRecord> iter = reader.iterator();
+            fields = iter.next().values();
+            if (fields.length < 2) {
```
VijayanB, line 96: you are assuming that header and records length will always be equal. I think you have to validate every record to verify whether they have at least two fields

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
```diff
@@ -0,0 +1,181 @@
...
+            .addSettingsUpdateConsumer(Ip2GeoSettings.INDEXING_BULK_SIZE, newValue -> this.indexingBulkSize = newValue);
+    }
+
+    /**
+     * Update GeoIp data internal
+     *
+     * @param jobParameter
+     * @throws Exception
+     */
+    public void updateGeoIpData(final Datasource jobParameter) throws Exception {
```
VijayanB, line 67:
```suggestion
    public void updateGeoIpData(final Datasource dataSource) throws Exception {
```

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
heemin32: It should have been created. Otherwise it won't get updated. Anyway I think it is better to say `doesn't exist`. Will update.

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceTransportAction.java
heemin32: Removed the method.

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
heemin32: Added validation code

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -264,14 +268,16 @@ public Long getLockDurationSeconds() {
      */
     @Override
     public Double getJitter() {
-        return 5.0 / (schedule.getInterval() * 24 * 60);
+        return 5.0 / (schedule.getInterval() * 24.0 * 60.0);
```
VijayanB, line 271: can you create constant for numbers?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -264,14 +268,16 @@ public Long getLockDurationSeconds() {
...
 
     /**
      * Enable auto update of GeoIP data
      */
     public void enable() {
-        enabledTime = Instant.now();
+        if (isEnabled == false) {
+            enabledTime = Instant.now().truncatedTo(ChronoUnit.MILLIS);
+        }
         isEnabled = true;
```
VijayanB, line 281:
```suggestion
        if (isEnabled == true) {
            return;
        }
        enabledTime = Instant.now().truncatedTo(ChronoUnit.MILLIS);
        isEnabled = true;
```

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceTransportAction.java
```diff
@@ -52,138 +41,96 @@
...
-                    if (e instanceof VersionConflictEngineException) {
-                        listener.onFailure(
-                            new ResourceAlreadyExistsException("datasource [{}] already exists", request.getDatasourceName())
-                        );
-                    } else {
-                        listener.onFailure(e);
-                    }
-                }
-            });
+            client.index(indexRequest, getIndexResponseListener(datasource, listener));
```
jmazanec15, line 80: Will this implicitly create the JOB_INDEX_NAME index? Do we want to explicitly control the creation of the index so that we explicitly set mapping as well as other settings?

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
```diff
@@ -41,49 +41,53 @@
...
- * Helper class for GeoIp data
+ * Facade class for GeoIp data
  */
 @Log4j2
-public class GeoIpDataHelper {
+public class GeoIpDataFacade {
     private static final String IP_RANGE_FIELD_NAME = "_cidr";
     private static final String DATA_FIELD_NAME = "_data";
     private static final Tuple<String, Integer> INDEX_SETTING_NUM_OF_SHARDS = new Tuple<>("index.number_of_shards", 1);
     private static final Tuple<String, String> INDEX_SETTING_AUTO_EXPAND_REPLICAS = new Tuple<>("index.auto_expand_replicas", "0-all");
```
jmazanec15, line 66: Not changed in this IP but how can user prevent auto_expanding to all replicas?

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
```diff
@@ -351,4 +344,13 @@ public static void putGeoData(
         client.admin().indices().prepareRefresh(indexName).execute().actionGet(timeout);
         client.admin().indices().prepareForceMerge(indexName).setMaxNumSegments(1).execute().actionGet(timeout);
     }
+
+    public AcknowledgedResponse deleteIndex(final String index) {
```
jmazanec15, line 348: Is this to delete any index? Should there be some restriction based on index name?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
```diff
@@ -118,161 +106,50 @@ public void runJob(final ScheduledJobParameter jobParameter, final JobExecutionC
...
-        if (!DatasourceState.AVAILABLE.equals(jobParameter.getState())) {
-            log.error("Invalid datasource state. Expecting {} but received {}", DatasourceState.AVAILABLE, jobParameter.getState());
-            jobParameter.disable();
-            jobParameter.getUpdateStats().setLastFailedAt(Instant.now());
-            DatasourceHelper.updateDatasource(client, jobParameter, timeout);
+    @VisibleForTesting
+    protected void updateDatasource(final ScheduledJobParameter jobParameter) throws IOException {
+        Datasource datasource = datasourceFacade.getDatasource(jobParameter.getName());
+        if (datasource == null) {
+            log.info("Datasource[{}] does not exist", jobParameter.getName());
```
jmazanec15, line 133: Should this throw an exception?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
```diff
@@ -118,161 +106,50 @@ public void runJob(final ScheduledJobParameter jobParameter, final JobExecutionC
...
-        URL url = new URL(jobParameter.getEndpoint());
-        DatasourceManifest manifest = DatasourceManifest.Builder.build(url);
-
-        if (skipUpdate(jobParameter, manifest)) {
-            log.info("Skipping GeoIP database update. Update is not required for {}", jobParameter.getId());
-            jobParameter.getUpdateStats().setLastSkippedAt(Instant.now());
-            DatasourceHelper.updateDatasource(client, jobParameter, timeout);
+        if (!DatasourceState.AVAILABLE.equals(datasource.getState())) {
+            log.error("Invalid datasource state. Expecting {} but received {}", DatasourceState.AVAILABLE, datasource.getState());
+            datasource.disable();
```
jmazanec15, line 139: Why is this being disabled?

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceTransportAction.java
heemin32: This index is controlled by job scheduler framework. They create the index accordingly. 
See. https://github.com/opensearch-project/geospatial/blob/feature/ip2geo/src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceExtension.java#L34

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
heemin32: User should not change it. It should be always auto expand so that ingestion latency can be fast enough.

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
heemin32: Let me add a condition check on index name.

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
heemin32: It can happen in normal use case. Delete request can be received while update task being waiting on queue for other update task to complete. When its turn come, it found that the datasource is deleted already so it should stop the update process.

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
heemin32: Otherwise, the next update will be scheduled but we don't want that to happen as it will fail again anyway.

src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
```diff
@@ -206,14 +169,45 @@ public void onFailure(final Exception e) {
...
+            return ipToGeoData;
+        }
+
+        filteredGeoData = new HashMap<>();
+        for (String property : this.properties) {
+            if (property.equals(PROPERTY_IP)) {
+                filteredGeoData.put(PROPERTY_IP, ip);
+            } else {
+                filteredGeoData.put(property, ipToGeoData.get(property));
+            }
```
VijayanB, line 197: This can be simplified. 
1. filteredGeoData = properties.stream().filter("not ip").collect(Collectors.toMap(...)
2. if properties contains "ip", filteredGeoData.put(PROPERTY_IP, ip);

heemin32: I am not convinced that suggested code is simpler than the original code..

VijayanB: What you were doing is imperative style. The suggested code is declarative. Declarative style conveys your logic better than existing way of mixing both logic in conditional statements.

heemin32: Fair enough.

heemin32:
@jmazanec15 @VijayanB Can I get another round of look at this PR?

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceTransportAction.java
jmazanec15: Above is our plugin. But I think we might need to create the index itself. In index management, they create the index on their own: https://github.com/opensearch-project/index-management/blob/2.7/src/main/kotlin/org/opensearch/indexmanagement/IndexManagementIndices.kt#L60

heemin32: I thought job scheduler will create index for us but seems it is not. Let me create the index explicitly in a separate PR as current PR is already too big to add another change here.

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Determine if update is needed or not
+     *
+     * Update is needed when all following conditions are met
+     * 1. MD5 hash value in datasource is different with MD5 hash value in manifest
+     * 2. updatedAt value in datasource is before updateAt value in manifest
+     *
+     * @param datasource
```
VijayanB, line 195: Your comments suggest that your method is shouldUpdate but the logic is skipUpdate. Can you keep it as consistency?

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
```diff
@@ -101,11 +107,11 @@ public static void createIndexIfNotExists(
...
      * @return String representing datasource database index mapping
      */
-    private static String getIndexMapping() {
+    private String getIndexMapping() {
         try {
-            try (InputStream is = DatasourceHelper.class.getResourceAsStream("/mappings/ip2geo_datasource.json")) {
+            try (InputStream is = DatasourceFacade.class.getResourceAsStream("/mappings/ip2geo_datasource.json")) {
                 try (BufferedReader reader = new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8))) {
-                    return reader.lines().collect(Collectors.joining());
+                    return reader.lines().map(line -> line.trim()).collect(Collectors.joining());
```
VijayanB, line 114:
```suggestion
                    return reader.lines().map(String::trim).collect(Collectors.joining());
```

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
```diff
@@ -351,4 +346,20 @@ public static void putGeoData(
         client.admin().indices().prepareRefresh(indexName).execute().actionGet(timeout);
         client.admin().indices().prepareForceMerge(indexName).setMaxNumSegments(1).execute().actionGet(timeout);
     }
+
+    public AcknowledgedResponse deleteIp2GeoDataIndex(final String index) {
+        if (index.startsWith(IP2GEO_DATA_INDEX_NAME_PREFIX) == false) {
```
VijayanB, line 351: nit: index null check?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -306,6 +316,11 @@ private String indexNameFor(final long suffix) {
         return String.format(Locale.ROOT, "%s.%s.%d", IP2GEO_DATA_INDEX_NAME_PREFIX, id, suffix);
     }
 
+    /**
+     * Tell if current datasource is expired or not
```
VijayanB, line 320: nit: may be Checks if datasource is expired or not

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
VijayanB: Can you update it in doc?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        Instant startTime = Instant.now();
+        String indexName = setupIndex(manifest, datasource);
+        String[] header;
+        List<String> fieldsToStore;
+        try (CSVParser reader = geoIpDataFacade.getDatabaseReader(manifest)) {
+            header = validateAndReturnHeader(reader.iterator()).values();
+            fieldsToStore = Arrays.asList(header).subList(1, header.length);
+            if (!datasource.isCompatible(fieldsToStore)) {
```
VijayanB, line 79: In some places i see x ==false is used and in another places !x is used. Shall we keep this consistent?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
```diff
@@ -0,0 +1,209 @@
...
+                try {
+                    if (geoIpDataFacade.deleteIp2GeoDataIndex(index).isAcknowledged()) {
+                        deletedIndices.add(index);
+                    } else {
+                        log.error("Failed to delete an index [{}]", index);
+                    }
+                } catch (Exception e) {
+                    log.error("Failed to delete an index [{}]", index, e);
+                }
+            }
```
VijayanB, line 120: nit: can you move this to private method?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Validate the next csv record as header
+     * This method move iterator one step forward
+     *
+     * @param iterator the csv record iterator
+     * @return next CSVRecord
+     */
+    private CSVRecord validateAndReturnHeader(Iterator<CSVRecord> iterator) {
```
VijayanB, line 137: nit: getHeadersIfValid

heemin32: getHeadersIfValid doesn't sounds right. Changed it as
```
CSVRecord headerLine = reader.iterator().next();
header = validateHeader(headerLine).values();
```

```
private CSVRecord validateHeader(CSVRecord header) {
    if (header == null) {
        throw new OpenSearchException("geoip database is empty");
    }
    if (header.values().length < 2) {
        throw new OpenSearchException("geoip database should have at least two fields");
    }
    return header;
}
```

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
heemin32: Done

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
heemin32: Done

src/main/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacade.java
heemin32: Done

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
heemin32: Done

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceRunner.java
heemin32: Done

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceUpdateService.java
heemin32: Done

heemin32:
@VijayanB @jmazanec15 Any more comments?

jmazanec15:
LGTM

VijayanB (approved):

PR #284, opened 2023-05-04
Update index option

heemin32:
### Description
1. Make geodata index as hidden
2. Make geodata index as read only allow delete after creation is done
3. Refresh datasource index immediately after update
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/test/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacadeTests.java
+++ b/src/test/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacadeTests.java
@@ -42,6 +42,7 @@ import org.opensearch.action.admin.indices.create.CreateIndexRequest;
 import org.opensearch.action.admin.indices.delete.DeleteIndexRequest;
 import org.opensearch.action.admin.indices.forcemerge.ForceMergeRequest;
 import org.opensearch.action.admin.indices.refresh.RefreshRequest;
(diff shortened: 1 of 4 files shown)


src/test/java/org/opensearch/geospatial/ip2geo/common/GeoIpDataFacadeTests.java
```diff
@@ -178,6 +181,12 @@ public void testPutGeoIpData() throws Exception {
                 assertEquals(index, request.indices()[0]);
                 assertEquals(1, request.maxNumSegments());
                 return null;
+            } else if (actionRequest instanceof UpdateSettingsRequest) {
+                UpdateSettingsRequest request = (UpdateSettingsRequest) actionRequest;
+                assertEquals(1, request.indices().length);
+                assertEquals(index, request.indices()[0]);
+                assertEquals("true", request.settings().get("index.blocks.read_only_allow_delete"));
```
martin-gaievski, line 188: Do you know why settings are returned as string despite the actual data type? Say for this one we have defined it as Boolean, but expected value is string true", not the boolean `true`.

heemin32: Yes.
```
public String get(String setting) {
        return toString(settings.get(setting));
    }
```

Let me use getAsBoolean instead.

junqiu-lei (approved):

martin-gaievski (approved):

PR #290, opened 2023-05-05
Add wrapper class of job scheduler lock service

heemin32:
### Description
Add wrapper class of job scheduler lock service
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial.ip2geo.common;
+
+import static org.opensearch.geospatial.ip2geo.jobscheduler.DatasourceExtension.JOB_INDEX_NAME;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.opensearch.action.ActionListener;
+import org.opensearch.client.Client;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.inject.Inject;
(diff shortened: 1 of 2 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
```diff
@@ -0,0 +1,101 @@
...
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.jobscheduler.spi.LockModel;
+import org.opensearch.jobscheduler.spi.utils.LockService;
+
+/**
+ * A wrapper of job scheduler's lock service for datasource
+ */
+public class Ip2GeoLockService {
```
martin-gaievski, line 28: Are you planning to add an interface for the service that this class will implement?

martin-gaievski (commented):
What's the client for the new service, do you plan another PR for this?

heemin32:
> What's the client for the new service, do you plan another PR for this?

Yes. Didn't include the client part so that each PR can be small.

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
heemin32: I don't have such plan.

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
```diff
@@ -0,0 +1,101 @@
...
+    private final ClusterService clusterService;
+    private final Client client;
+    private final LockService lockService;
+
+    /**
+     * Default constructor
```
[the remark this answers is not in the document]
heemin32: If there is constructor with argument, compiler does not create no argument constructor by default. Therefore, we don't need to make private no argument constructor.

martin-gaievski (approved):
Thank you for addressing all my comments, LGTM

junqiu-lei (approved):

PR #291, opened 2023-05-05
Implements delete datasource API

heemin32:
### Description
Implements delete datasource API

 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequest.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequest.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial.ip2geo.action;
+
+import java.io.IOException;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
+
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.common.io.stream.StreamInput;
+import org.opensearch.common.io.stream.StreamOutput;
+
+/**
+ * GeoIP datasource delete request
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+public class DeleteDatasourceRequest extends ActionRequest {
+    /**
+     * @param name the datasource name
+     * @return the datasource name
+     */
+    private String name;
+
+    /**
+     * Constructor
+     *
+     * @param in the stream input
+     * @throws IOException IOException
+     */
+    public DeleteDatasourceRequest(final StreamInput in) throws IOException {
+        super(in);
+        this.name = in.readString();
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        ActionRequestValidationException errors = null;
+        if (name == null || name.isBlank()) {
+            errors = new ActionRequestValidationException();
+            errors.addValidationError("Datasource name should not be empty");
+        }
+        return errors;
+    }
+
+    @Override
+    public void writeTo(final StreamOutput out) throws IOException {
+        super.writeTo(out);
+        out.writeString(name);
+    }
+}
--- a/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial.ip2geo.action;
+
+import java.io.IOException;
+
+import lombok.extern.log4j.Log4j2;
+
+import org.opensearch.OpenSearchException;
+import org.opensearch.ResourceNotFoundException;
+import org.opensearch.action.ActionListener;
+import org.opensearch.action.support.ActionFilters;
+import org.opensearch.action.support.HandledTransportAction;
+import org.opensearch.action.support.master.AcknowledgedResponse;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.geospatial.annotation.VisibleForTesting;
+import org.opensearch.geospatial.ip2geo.common.DatasourceFacade;
+import org.opensearch.geospatial.ip2geo.common.DatasourceState;
+import org.opensearch.geospatial.ip2geo.common.Ip2GeoLockService;
+import org.opensearch.geospatial.ip2geo.jobscheduler.Datasource;
+import org.opensearch.geospatial.ip2geo.processor.Ip2GeoProcessor;
+import org.opensearch.ingest.IngestMetadata;
+import org.opensearch.ingest.IngestService;
+import org.opensearch.tasks.Task;
+import org.opensearch.transport.TransportService;
+
+/**
+ * Transport action to delete datasource
+ */
+@Log4j2
+public class DeleteDatasourceTransportAction extends HandledTransportAction<DeleteDatasourceRequest, AcknowledgedResponse> {
+    private static final long LOCK_DURATION_IN_SECONDS = 300l;
+    private final Ip2GeoLockService lockService;
+    private final IngestService ingestService;
+    private final DatasourceFacade datasourceFacade;
+
+    /**
+     * Constructor
+     * @param transportService the transport service
+     * @param actionFilters the action filters
+     * @param lockService the lock service
+     * @param ingestService the ingest service
+     * @param datasourceFacade the datasource facade
+     */
+    @Inject
+    public DeleteDatasourceTransportAction(
+        final TransportService transportService,
+        final ActionFilters actionFilters,
+        final Ip2GeoLockService lockService,
+        final IngestService ingestService,
+        final DatasourceFacade datasourceFacade
+    ) {
+        super(DeleteDatasourceAction.NAME, transportService, actionFilters, DeleteDatasourceRequest::new);
+        this.lockService = lockService;
+        this.ingestService = ingestService;
+        this.datasourceFacade = datasourceFacade;
+    }
+
+    /**
+     * We delete datasource regardless of its state as long as we can acquire a lock
+     *
+     * @param task the task
+     * @param request the request
+     * @param listener the listener
+     */
+    @Override
+    protected void doExecute(final Task task, final DeleteDatasourceRequest request, final ActionListener<AcknowledgedResponse> listener) {
+        lockService.acquireLock(request.getName(), LOCK_DURATION_IN_SECONDS, ActionListener.wrap(lock -> {
+            if (lock == null) {
+                listener.onFailure(new OpenSearchException("another processor is holding a lock on the resource. Try again later"));
+                return;
+            }
+            try {
+                deleteDatasource(request.getName());
+                listener.onResponse(new AcknowledgedResponse(true));
+            } catch (Exception e) {
+                listener.onFailure(e);
+            } finally {
+                lockService.releaseLock(
+                    lock,
+                    ActionListener.wrap(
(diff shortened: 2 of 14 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
```diff
@@ -0,0 +1,132 @@
...
+    private void setDatasourceStateAsDeleting(final Datasource datasource) throws IOException {
+        if (isSafeToDelete(datasource) == false) {
+            new OpenSearchException("datasource is being used by one of processors");
+        }
+
+        DatasourceState previousState = datasource.getState();
+        datasource.setState(DatasourceState.DELETING);
+        datasourceFacade.updateDatasource(datasource);
+
+        // Check again as processor might just have been created.
```
jmazanec15, line 111: Is there not a write lock that can be grabbed to delete the datasource?

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
```diff
@@ -0,0 +1,132 @@
...
+            new ResourceNotFoundException("no such datasource exist");
+        }
+
+        setDatasourceStateAsDeleting(datasource);
+        datasourceFacade.deleteDatasource(datasource);
+    }
+
+    private void setDatasourceStateAsDeleting(final Datasource datasource) throws IOException {
+        if (isSafeToDelete(datasource) == false) {
+            new OpenSearchException("datasource is being used by one of processors");
```
jmazanec15, line 104: Why doesnt this have throw before it?

src/main/java/org/opensearch/geospatial/ip2geo/common/DatasourceFacade.java
```diff
@@ -137,6 +141,34 @@ public IndexResponse updateDatasource(final Datasource datasource) throws IOExce
...
+     */
+    public void deleteDatasource(final Datasource datasource) {
+        if (client.admin()
+            .indices()
+            .prepareDelete(datasource.getIndices().toArray(new String[0]))
+            .setIndicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN)
+            .execute()
+            .actionGet(clusterSettings.get(Ip2GeoSettings.TIMEOUT))
+            .isAcknowledged() == false) {
+            throw new OpenSearchException("failed to delete data[{}] in datasource", StringUtils.joinWith(",", datasource.getIndices()));
```
jmazanec15, line 158: Why generic OpenSearchException?

heemin32: There is no specific action that client can take on this error. So using generic OpenSearch exception.

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
heemin32: before means? A client will see the error message in response when this exception is thrown.

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
heemin32: Haven't implemented lock between create ip2geo processor and delete datasource.
Appreciate if you could share some idea on the write lock. If write lock involves interacting with index, it cannot be used during ip2geo processor creation phase because processor creation is a process which changes cluster state and while changing cluster state, we cannot call some or all of OpenSearch API like getting or searching document from an index.

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @param task the task
+     * @param request the request
+     * @param listener the listener
+     */
+    @Override
+    protected void doExecute(final Task task, final DeleteDatasourceRequest request, final ActionListener<AcknowledgedResponse> listener) {
+        lockService.acquireLock(request.getName(), LOCK_DURATION_IN_SECONDS, ActionListener.wrap(lock -> {
+            if (lock == null) {
+                listener.onFailure(new OpenSearchException("another processor is holing a lock on the resource. Try again later"));
```
jmazanec15, line 72: holing - > holding

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
```diff
@@ -0,0 +1,132 @@
...
+                return;
+            }
+            try {
+                deleteDatasource(request.getName());
```
jmazanec15, line 76: Why not pass listener to this method?

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
jmazanec15: When does it get thrown?

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
jmazanec15: hmmm yes Im not sure what the lock would be. I would think there would be something with the JobScheduler LockService.

heemin32: Lock with job scheduler utilize opensearch index which cannot be used in processor creation time.

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
heemin32: If a use trying to delete datasource using DELETE api, it check if there is a processor associated with the datasource. If there is, then the exception is thrown and returned back to the user.

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
heemin32: There is no async call in the method. Not much difference between current one and passing the listener down the sub method but only increase complexity.

src/main/java/org/opensearch/geospatial/ip2geo/action/RestDeleteDatasourceHandler.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final String ACTION_NAME = "ip2geo_datasource_delete";
+
+    @Override
+    public String getName() {
+        return ACTION_NAME;
+    }
+
+    @Override
+    protected RestChannelConsumer prepareRequest(final RestRequest request, final NodeClient client) {
+        final String name = request.param("name");
```
VijayanB, line 32: nit: can you move it as a constant?

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
heemin32: Yes. The thing is, we didn't want to add any custom metadata during design review meeting as it can cause a headache in the future due to many restriction on maintaining backward compatibility.

src/main/java/org/opensearch/geospatial/ip2geo/common/DatasourceFacade.java
VijayanB, line 158: nit: Why not use String.join method. My recommendation is to prefer jdk over external dependency

src/main/java/org/opensearch/geospatial/ip2geo/common/DatasourceFacade.java
```diff
@@ -137,6 +141,34 @@ public IndexResponse updateDatasource(final Datasource datasource) throws IOExce
...
+        }
+        DeleteResponse response = client.prepareDelete()
+            .setIndex(DatasourceExtension.JOB_INDEX_NAME)
+            .setId(datasource.getName())
+            .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
+            .execute()
+            .actionGet(clusterSettings.get(Ip2GeoSettings.TIMEOUT));
+
+        if (response.status().equals(RestStatus.OK) == false && response.status().equals(RestStatus.NOT_FOUND) == false) {
+            throw new OpenSearchException("failed to delete the datasource");
```
VijayanB, line 168: nit: can you log response here ?

src/main/java/org/opensearch/geospatial/ip2geo/common/DatasourceFacade.java
```diff
@@ -137,6 +141,34 @@ public IndexResponse updateDatasource(final Datasource datasource) throws IOExce
...
+            throw new OpenSearchException("failed to delete data[{}] in datasource", StringUtils.joinWith(",", datasource.getIndices()));
+        }
+        DeleteResponse response = client.prepareDelete()
+            .setIndex(DatasourceExtension.JOB_INDEX_NAME)
+            .setId(datasource.getName())
+            .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
+            .execute()
+            .actionGet(clusterSettings.get(Ip2GeoSettings.TIMEOUT));
+
+        if (response.status().equals(RestStatus.OK) == false && response.status().equals(RestStatus.NOT_FOUND) == false) {
```
VijayanB, line 167: nit: we should break this if and provide error message accordingly. ex: 404 means, datasource doesn't exist. and for others we can say failed to delete datasource

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
heemin32: Talked offline. Missed to add `throw`

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequest.java
```diff
@@ -0,0 +1,65 @@
...
+     * @return the datasource name
+     */
+    private String name;
+
+    /**
+     * Constructor
+     *
+     * @param name list of datasource name
+     */
+    public DeleteDatasourceRequest(final String name) {
```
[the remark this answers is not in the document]
heemin32: For this simple one, I think having explicit constructor doesn't harm much. Let me keep it.

src/main/java/org/opensearch/geospatial/ip2geo/common/DatasourceFacade.java
```diff
@@ -164,6 +166,38 @@ public void putDatasource(final Datasource datasource, final ActionListener list
...
+            .execute()
+            .actionGet(clusterSettings.get(Ip2GeoSettings.TIMEOUT));
+
+        if (response.status().equals(RestStatus.OK)) {
+            log.info("deleted datasource[{}] successfully", datasource.getName());
+        } else if (response.status().equals(RestStatus.NOT_FOUND)) {
+            throw new ResourceNotFoundException("datasource[{}] does not exist", datasource.getName());
+        } else {
+            throw new OpenSearchException("failed to delete datasource[{}] with status[{}]", datasource.getName(), response.status());
+        }
```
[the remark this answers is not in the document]
heemin32: I would like to keep the code as it. Having them in a single if/ifelse/else statement makes sense to me because they are all checking value on the same variable. reponse.status().

VijayanB (approved):

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
jmazanec15: Got it, might be something to consider in the future if this becomes a problem.

jmazanec15 (approved):

PR #305, opened 2023-05-12
Bug fix and refactoring of code

heemin32:
### Description
1. Bugfix: Ingest metadata can be null if there is no processor created
2. Refactoring: Moved private method to another class for better testing support
3. Refactoring: Set some private static final variable as public so that unit test can use it
4. Refactoring: Changed string value to static variable
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/test/java/org/opensearch/geospatial/ip2geo/Ip2GeoTestCase.java
+++ b/src/test/java/org/opensearch/geospatial/ip2geo/Ip2GeoTestCase.java
@@ -16,6 +16,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Locale;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


src/test/java/org/opensearch/geospatial/ip2geo/Ip2GeoTestCase.java
```diff
@@ -211,6 +216,28 @@ protected LockModel randomLockModel() {
         return lockModel;
     }
 
+    protected Ip2GeoProcessor randomIp2GeoProcessor(String datasourceName) {
```
martin-gaievski, line 219: why this method called randomGeoProcessor? My initial intuition was that we have few implementations and method will randomly choose one of it, but looks like we're returning same one on every call

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceRequest.java
```diff
@@ -36,8 +36,8 @@
 @Log4j2
 @EqualsAndHashCode
 public class PutDatasourceRequest extends AcknowledgedRequest<PutDatasourceRequest> {
-    private static final ParseField ENDPOINT_FIELD = new ParseField("endpoint");
-    private static final ParseField UPDATE_INTERVAL_IN_DAYS_FIELD = new ParseField("update_interval_in_days");
+    public static final ParseField ENDPOINT_FIELD = new ParseField("endpoint");
+    public static final ParseField UPDATE_INTERVAL_IN_DAYS_FIELD = new ParseField("update_interval_in_days");
```
martin-gaievski, line 40: why those two constants made public, seems we're not changing any code accessing it

src/test/java/org/opensearch/geospatial/ip2geo/Ip2GeoTestCase.java
heemin32: Random here means values that we are going to set in datasource. Except datasourceName, all other values are randomly generated.

src/main/java/org/opensearch/geospatial/ip2geo/action/PutDatasourceRequest.java
heemin32: Going to use it in integ test that I am going to raise in another PR.

heemin32: Will remove it from this PR..

martin-gaievski (approved):

junqiu-lei (approved):

PR #319, opened 2023-05-24
Remove jitter and move index setting from DatasourceFacade to Datasou…

heemin32:
### Description
Remove jitter and move index setting from DatasourceFacade to DatasourceExtension 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceExtension.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceExtension.java
@@ -5,6 +5,8 @@
 
 package org.opensearch.geospatial.ip2geo.jobscheduler;
 
+import java.util.Map;
+
 import org.opensearch.jobscheduler.spi.JobSchedulerExtension;
 import org.opensearch.jobscheduler.spi.ScheduledJobParser;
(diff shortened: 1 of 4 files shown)


naveentatikonda (approved):

junqiu-lei (approved):

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/DatasourceExtension.java
```diff
@@ -21,6 +23,20 @@ public class DatasourceExtension implements JobSchedulerExtension {
...
+     * We want it to expand to all replicas so that querying to this index can be done locally to reduce latency.
+     */
+    public static final Map<String, Object> INDEX_SETTING = Map.of(
+        "index.number_of_shards",
+        1,
+        "index.auto_expand_replicas",
+        "0-all",
+        "index.hidden",
+        true
+    );
```
martin-gaievski, line 39: I think we need constants for keys, previously each separate variable as like such constant, but we lost it with the map

heemin32: INDEX_SETTING itself is a constant so I didn't make value in it as another constant.

PR #328, opened 2023-06-01
Add restoring event listener

heemin32:
### Description
In the listener, we trigger a geoip data update
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
@@ -354,7 +354,15 @@ public class Datasource implements Writeable, ScheduledJobParameter {
      * @return Current index name of a datasource
      */
     public String currentIndexName() {
-        return isExpired() ? null : indexNameFor(database.updatedAt.toEpochMilli());
+        if (isExpired()) {
+            return null;
+        }
+
+        if (database.updatedAt == null) {
+            return null;
+        }
+
+        return indexNameFor(database.updatedAt.toEpochMilli());
     }
 
     /**
@@ -369,6 +377,14 @@ public class Datasource implements Writeable, ScheduledJobParameter {
 
     private String indexNameFor(final long suffix) {
         return String.format(Locale.ROOT, "%s.%s.%d", IP2GEO_DATA_INDEX_NAME_PREFIX, name, suffix);
+    }
+
+    /**
+     * Reset database so that it can be updated in next run regardless there is new update or not
+     */
+    public void resetDatabase() {
+        database.setUpdatedAt(null);
+        database.setSha256Hash(null);
     }
 
     /**
@@ -537,7 +553,7 @@ public class Datasource implements Writeable, ScheduledJobParameter {
         public void writeTo(final StreamOutput out) throws IOException {
             out.writeOptionalString(provider);
             out.writeOptionalString(sha256Hash);
-            out.writeOptionalVLong(updatedAt.toEpochMilli());
+            out.writeOptionalVLong(updatedAt == null ? null : updatedAt.toEpochMilli());
             out.writeOptionalVLong(validForInDays);
             out.writeOptionalStringCollection(fields);
         }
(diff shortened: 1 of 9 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -537,7 +553,7 @@ public Database(final StreamInput in) throws IOException {
         public void writeTo(final StreamOutput out) throws IOException {
             out.writeOptionalString(provider);
             out.writeOptionalString(sha256Hash);
-            out.writeOptionalVLong(updatedAt.toEpochMilli());
+            out.writeOptionalVLong(updatedAt == null ? null : updatedAt.toEpochMilli());
```
martin-gaievski, line 556: isn't the check for null already encapsulated into method if it does "optional"?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -640,10 +656,10 @@ public UpdateStats(final StreamInput in) throws IOException {
 
         @Override
         public void writeTo(final StreamOutput out) throws IOException {
-            out.writeOptionalVLong(lastSucceededAt.toEpochMilli());
+            out.writeOptionalVLong(lastSucceededAt == null ? null : lastSucceededAt.toEpochMilli());
```
martin-gaievski, line 659: same as for previous call to write optional, I think null value should be already handled

src/main/java/org/opensearch/geospatial/ip2geo/listener/Ip2GeoListener.java
```diff
@@ -0,0 +1,118 @@
...
+import org.opensearch.jobscheduler.spi.schedule.IntervalSchedule;
+import org.opensearch.threadpool.ThreadPool;
+
+@Log4j2
+public class Ip2GeoListener extends AbstractLifecycleComponent implements ClusterStateListener {
+    private static final int SCHEDULE_IN_MIN = 15;
+    private static final int DELAY_IN_MILLIS = 10000;
+    private final ClusterService clusterService;
+    private final ThreadPool threadPool;
+
```
martin-gaievski, line 37: nit: extra empty line

src/main/java/org/opensearch/geospatial/ip2geo/listener/Ip2GeoListener.java
```diff
@@ -0,0 +1,118 @@
...
+    private final DatasourceFacade datasourceFacade;
+
+    @Inject
```
martin-gaievski, line 40: curious, is this different/can be replaced by lombok's AllArgsConstructor?

src/main/java/org/opensearch/geospatial/ip2geo/listener/Ip2GeoListener.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    protected void doStop() {
+        clusterService.removeListener(this);
+    }
+
+    @Override
+    protected void doClose() throws IOException {
+
```
martin-gaievski, line 116: should we do same as on stop here?

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
heemin32: Here I call `.toEpochMillis()` method and will will throw NPE if I don't do null check.

src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
heemin32: Same for previous one. I call `.toEpochMillis()` method.

src/main/java/org/opensearch/geospatial/ip2geo/listener/Ip2GeoListener.java
heemin32: I think `doStop` is called always before `doClose()`. Removing listener in `doClose()` will have no effect.
Saw the same pattern in OpenSearch core.

martin-gaievski (approved):

junqiu-lei (approved):

PR #333, opened 2023-06-07
Add a field in datasource for current index name

heemin32:
### Description
Add a field in datasource for current index name and use current time as suffix for new index name.
This is needed to handle snapshot restoring in next PR where we will create a new index always and delete any restored index.

 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
@@ -89,6 +89,7 @@ public class Datasource implements Writeable, ScheduledJobParameter {
      */
     private static final ParseField ENDPOINT_FIELD = new ParseField("endpoint");
     private static final ParseField STATE_FIELD = new ParseField("state");
+    private static final ParseField CURRENT_INDEX_FIELD = new ParseField("current_index");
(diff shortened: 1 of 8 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/jobscheduler/Datasource.java
```diff
@@ -358,25 +374,18 @@ public String currentIndexName() {
...
-
-        return indexNameFor(database.updatedAt.toEpochMilli());
+        return currentIndex;
     }
 
     /**
-     * Index name for a given manifest
+     * New index name for a datasource with current time as suffix
+     *
+     * This method is not thread safe. It could generate same index depends on what Instant.now() returns.
```
junqiu-lei, line 383: Any potential issue here for not thread safe? Could we use random str instead of current time to mitigate the thread risk

heemin32: No issue as of current usage. I think I need to remove that comment. It is thread safe, just it does not guarantee the uniqueness of index name.

junqiu-lei (approved):

naveentatikonda (approved):

PR #335, opened 2023-06-09
Use bool query for array form of IPs and wrapping with try/catch

heemin32:
### Description
1. Instead of msearch for array form of IPs, use bool query for better performance.
2. IP property support is removed which add ip address in enriched data. This was added at first for feature parity with geoip processor but it cannot be supported in array form as we are using bool query now.
3. Added try/catch for every possible cases so that exception one impact to other requests in the same bulk request.
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
@@ -11,10 +11,9 @@ import static org.opensearch.ingest.ConfigurationUtils.readOptionalList;
 import static org.opensearch.ingest.ConfigurationUtils.readStringProperty;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
```diff
@@ -111,19 +107,23 @@ public Ip2GeoProcessor(
...
-        } else if (ip instanceof List) {
-            executeInternal(ingestDocument, handler, ((List<?>) ip));
-        } else {
-            throw new IllegalArgumentException("field [" + field + "] should contain only string or array of strings");
+            if (ip instanceof String) {
+                executeInternal(ingestDocument, handler, (String) ip);
+            } else if (ip instanceof List) {
+                executeInternal(ingestDocument, handler, ((List<?>) ip));
+            } else {
+                handler.accept(null, new IllegalArgumentException("field [" + field + "] should contain only string or array of strings"));
```
martin-gaievski, line 123: please use String.format for exception message

heemin32:
Build is failing because of refactoring done in OpenSearch core. It might get resolved once the change is deployed in docker repo.

martin-gaievski (approved):

naveentatikonda (approved):

PR #336, opened 2023-06-13
Remove IP2Geo processor validation

heemin32:
### Description
Cannot query index to get data to validate IP2Geo processor. Will add validation when we decide to store some of data in cluster state metadata.
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
@@ -4,8 +4,6 @@
  */
 package org.opensearch.geospatial.ip2geo.processor;
 
-import static org.opensearch.cluster.service.ClusterApplierService.CLUSTER_UPDATE_THREAD_NAME;
-import static org.opensearch.ingest.ConfigurationUtils.newConfigurationException;
 import static org.opensearch.ingest.ConfigurationUtils.readBooleanProperty;
 import static org.opensearch.ingest.ConfigurationUtils.readOptionalList;
 import static org.opensearch.ingest.ConfigurationUtils.readStringProperty;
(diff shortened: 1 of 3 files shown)


junqiu-lei (approved):

src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
```diff
@@ -174,6 +173,10 @@ public void onResponse(final Datasource datasource) {
 
             @Override
             public void onFailure(final Exception e) {
+                if (e instanceof IndexNotFoundException) {
```
martin-gaievski, line 176: is check on exception class enough here, should we also check on a message to minimize chance of false positive?

src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
```diff
@@ -153,8 +152,8 @@ protected void executeInternal(
         datasourceFacade.getDatasource(datasourceName, new ActionListener<>() {
             @Override
             public void onResponse(final Datasource datasource) {
-                if (datasource == null) {
-                    handler.accept(null, new IllegalStateException("datasource does not exist"));
+                if (datasource == null || DatasourceState.AVAILABLE.equals(datasource.getState()) == false) {
```
martin-gaievski, line 155: could you please put this check into a method with a name that reflects condition that we're checking here, it's a bit hard to read? Method can be named something like "isDataSourceUnavailable"

src/main/java/org/opensearch/geospatial/ip2geo/processor/Ip2GeoProcessor.java
heemin32: Check on exception class is enough. No other cases. It is only for error message display so not a big issue even if there is false positive.

naveentatikonda (approved):

PR #337, opened 2023-06-13
Run update/delete request in a new thread

heemin32:
### Description
This is not to block transport thread
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/action/UpdateDatasourceTransportAction.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/action/UpdateDatasourceTransportAction.java
@@ -31,6 +31,7 @@ import org.opensearch.geospatial.ip2geo.jobscheduler.DatasourceTask;
 import org.opensearch.geospatial.ip2geo.jobscheduler.DatasourceUpdateService;
 import org.opensearch.jobscheduler.spi.schedule.IntervalSchedule;
 import org.opensearch.tasks.Task;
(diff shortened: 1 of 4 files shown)


junqiu-lei (approved):

src/main/java/org/opensearch/geospatial/ip2geo/action/UpdateDatasourceTransportAction.java
```diff
@@ -82,16 +86,23 @@ protected void doExecute(final Task task, final UpdateDatasourceRequest request,
...
-                }
-                validate(request, datasource);
-                updateIfChanged(request, datasource);
-                lockService.releaseLock(lock);
-                listener.onResponse(new AcknowledgedResponse(true));
+                // TODO: makes every sub-methods as async call to avoid using a thread in generic pool
+                threadPool.generic().submit(() -> {
+                    try {
+                        Datasource datasource = datasourceFacade.getDatasource(request.getName());
+                        if (datasource == null) {
```
martin-gaievski, line 94: if this check costly? If check is failed we can skip the overhead of spinning new thread worker? I think it's better f we can check this before submitting a new job.

heemin32: Not costly but I will work on it as a follow up action item. It is commented as TODO.

naveentatikonda (approved):

PR #338, opened 2023-06-21
Added a cache to store datasource metadata

heemin32:
### Description
Added a cache to store datasource metadata
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/cache/Ip2GeoCache.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/cache/Ip2GeoCache.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial.ip2geo.cache;
+
+import java.io.IOException;
+import java.time.Instant;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
+
+import org.opensearch.common.xcontent.LoggingDeprecationHandler;
+import org.opensearch.common.xcontent.XContentType;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.geospatial.ip2geo.common.DatasourceFacade;
+import org.opensearch.geospatial.ip2geo.common.DatasourceState;
+import org.opensearch.geospatial.ip2geo.jobscheduler.Datasource;
+import org.opensearch.index.engine.Engine;
+import org.opensearch.index.shard.IndexingOperationListener;
+import org.opensearch.index.shard.ShardId;
+
+@Log4j2
+public class Ip2GeoCache implements IndexingOperationListener {
+    private final DatasourceFacade datasourceFacade;
+    private Map<String, DatasourceMetadata> data;
+
+    public Ip2GeoCache(final DatasourceFacade datasourceFacade) {
+        this.datasourceFacade = datasourceFacade;
+    }
+
+    public String getIndexName(final String datasourceName) {
+        return getData().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getIndexName();
+    }
+
+    public boolean isExpired(final String datasourceName) {
+        return getData().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getExpirationDate().isBefore(Instant.now());
+    }
+
+    public boolean has(final String datasourceName) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/cache/Ip2GeoCache.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private void put(final Datasource datasource) {
+        DatasourceMetadata metadata = new DatasourceMetadata(datasource);
+        getData().put(datasource.getName(), metadata);
+    }
+
+    private void remove(final String datasourceName) {
+        getData().remove(datasourceName);
+    }
```
junqiu-lei, line 76: The put and remove methods directly access the data map without any synchronization, shall we ensure that only one thread can modify the data at one time?

heemin32: It is a concurrent hash map and it will handle concurrent behavior.

junqiu-lei (approved):

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
```diff
@@ -96,19 +105,26 @@ public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings sett
 
     @Override
     public Map<String, Processor.Factory> getProcessors(Processor.Parameters parameters) {
+        this.datasourceFacade = new DatasourceFacade(parameters.client, parameters.ingestService.getClusterService());
+        this.geoIpDataFacade = new GeoIpDataFacade(parameters.ingestService.getClusterService(), parameters.client);
+        this.ip2GeoCache = new Ip2GeoCache(datasourceFacade);
```
martin-gaievski, line 110: is it possible to inject those three service refs instead? this will make testing easier as we can inject mocks. In addition to that it's easy to adopt this code for later changes in case we'll have multiple implementation(s) for any of the services

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
```diff
@@ -96,19 +105,26 @@ public Collection<SystemIndexDescriptor> getSystemIndexDescriptors(Settings sett
...
-                    new GeoIpDataFacade(parameters.ingestService.getClusterService(), parameters.client)
-                )
+                new Ip2GeoProcessor.Factory(parameters.ingestService, datasourceFacade, geoIpDataFacade, ip2GeoCache)
             )
             .immutableMap();
     }
 
+    @Override
+    public void onIndexModule(IndexModule indexModule) {
+        if (indexModule.getIndex().getName().equals(DatasourceExtension.JOB_INDEX_NAME)) {
```
martin-gaievski, line 122:
```suggestion
        if (DatasourceExtension.JOB_INDEX_NAME.equals(indexModule.getIndex().getName())) {
```

src/main/java/org/opensearch/geospatial/ip2geo/cache/Ip2GeoCache.java
```diff
@@ -0,0 +1,120 @@
...
+    public Ip2GeoCache(final DatasourceFacade datasourceFacade) {
+        this.datasourceFacade = datasourceFacade;
+    }
+
+    public String getIndexName(final String datasourceName) {
+        return getData().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getIndexName();
+    }
+
+    public boolean isExpired(final String datasourceName) {
+        return getData().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getExpirationDate().isBefore(Instant.now());
```
martin-gaievski, line 41: why not to fail fast instead of returning some data based on empty_metadata? If we pretend that datasource exists then it may lead to errors on the client side if we return something from "isExpired" or "getIndexName"

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
heemin32: It is not possible to use dependency injection in processor factory creation. There is one option to avoid this parameter passing by using transport action inside processor. However, it will increase latency and code complexity.

src/main/java/org/opensearch/geospatial/ip2geo/cache/Ip2GeoCache.java
heemin32: I choose not to throw exception but return default value. There is no risk of pretending datasource exist by this method because this method should not be used to check if datasource exist or not. There is another method to check that. This just simply the code with low risk. If you still think throwing exception is better, I can make a change.

martin-gaievski (approved):

PR #340, opened 2023-06-22
Add geo data cache

heemin32:
### Description
Add geo data cache
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
@@ -7,16 +7,24 @@ package org.opensearch.geospatial.ip2geo.dao;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
 
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.extern.log4j.Log4j2;
 
+import org.opensearch.common.cache.Cache;
+import org.opensearch.common.cache.CacheBuilder;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.XContentType;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.geospatial.annotation.VisibleForTesting;
 import org.opensearch.geospatial.ip2geo.common.DatasourceState;
 import org.opensearch.geospatial.ip2geo.jobscheduler.Datasource;
 import org.opensearch.index.engine.Engine;
@@ -117,6 +125,67 @@ public class Ip2GeoCachedDao implements IndexingOperationListener {
             this.indexName = datasource.currentIndexName();
             this.expirationDate = datasource.expirationDay();
             this.state = datasource.getState();
+        }
+    }
+
+    /**
+     * Cache to hold geo data
+     *
+     * GeoData in an index in immutable. Therefore, invalidation is not needed.
+     */
+    @VisibleForTesting
+    protected static class GeoDataCache {
+        private Cache<CacheKey, Map<String, Object>> cache;
+
+        public GeoDataCache(final long maxSize) {
+            if (maxSize < 0) {
+                throw new IllegalArgumentException("ip2geo max cache size must be 0 or greater");
+            }
+            this.cache = CacheBuilder.<CacheKey, Map<String, Object>>builder().setMaximumWeight(maxSize).build();
+        }
+
+        public Map<String, Object> putIfAbsent(
+            final String indexName,
+            final String ip,
+            final Function<String, Map<String, Object>> retrieveFunction
+        ) throws ExecutionException {
+            CacheKey cacheKey = new CacheKey(indexName, ip);
+            return cache.computeIfAbsent(cacheKey, key -> retrieveFunction.apply(key.ip));
+        }
+
+        public Map<String, Object> get(final String indexName, final String ip) {
+            return cache.get(new CacheKey(indexName, ip));
+        }
+
+        /**
+         * Create a new cache with give size and replace existing cache
+         *
+         * Try to populate the existing value from previous cache to the new cache in best effort
+         *
+         * @param maxSize
+         */
+        public void updateMaxSize(final long maxSize) {
+            if (maxSize < 0) {
+                throw new IllegalArgumentException("ip2geo max cache size must be 0 or greater");
+            }
+            Cache<CacheKey, Map<String, Object>> temp = CacheBuilder.<CacheKey, Map<String, Object>>builder()
+                .setMaximumWeight(maxSize)
+                .build();
+            int count = 0;
+            Iterator<CacheKey> it = cache.keys().iterator();
+            while (it.hasNext() && count < maxSize) {
+                CacheKey key = it.next();
+                temp.put(key, cache.get(key));
+                count++;
+            }
+            cache = temp;
+        }
+
+        @AllArgsConstructor
+        @EqualsAndHashCode
+        private static class CacheKey {
+            private final String indexName;
+            private final String ip;
         }
     }
 }
--- a/src/test/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDaoTests.java
+++ b/src/test/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDaoTests.java
@@ -9,12 +9,16 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
```diff
@@ -0,0 +1,79 @@
...
+    public Map<String, Object> get(final String datasourceName, final InetAddress ip) {
+        return cache.get(new CacheKey(datasourceName, ip));
+    }
+
+    /**
+     * Create a new cache with give size and replace existing cache
+     *
+     * @param maxSize
+     */
+    public void updateMaxSize(final long maxSize) {
```
VijayanB, line 51: This will clear existing cache. is this expected?

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
```diff
@@ -0,0 +1,79 @@
...
+import org.opensearch.common.cache.Cache;
+import org.opensearch.common.cache.CacheBuilder;
+
+/**
+ * Cache to hold geo data
+ */
+public class GeoDataCache {
+    private Cache<CacheKey, Map<String, Object>> cache;
+
+    public GeoDataCache(final long maxSize) {
```
VijayanB, line 26: Shouldn't this be singleton?

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
heemin32: Yes. On a second thought, let me try to copy the existing value from old to new cache.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
heemin32: That make sense.

heemin32: Cannot make it as a singleton as the maxSize property is coming from the caller.

VijayanB: Does it mean you have multiple cache per cluster?

heemin32: There will be single cache per node. Any suggestion on how you want to make it as singleton? I would leave the code as it is and let the user of this class to decide. Because same can happen for many other classes as well which are singleton but not protection on creating multiple instance.

junqiu-lei (approved):

navneet1v:
@heemin32 attach an issue with this PR.


Also add some information about this PR, what it contains

heemin32:
> @heemin32 attach an issue with this PR.
> 
> Also add some information about this PR, what it contains

There is no issue specific to this PR. I don't want to add the IP2Geo issue as there so many PR and it will spam the issue page. It contains a class implementing a cache...

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param maxSize
+     */
+    public void updateMaxSize(final long maxSize) {
+        if (maxSize < 0) {
+            throw new IllegalArgumentException("ip2geo max cache size must be 0 or greater");
+        }
+        Cache<CacheKey, Map<String, Object>> temp = CacheBuilder.<CacheKey, Map<String, Object>>builder().setMaximumWeight(maxSize).build();
+        int count = 0;
+        Iterator<GeoDataCache.CacheKey> it = cache.keys().iterator();
+        while (it.hasNext() && count < maxSize) {
```
navneet1v, line 61: This is an interesting condition. We are providing the update to max size, and if new size if less than the old size we are copying less data. This adds more doubt in my mind why we are even providing this functionality.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
```diff
@@ -0,0 +1,75 @@
...
+            GeoDataCache.CacheKey key = it.next();
+            temp.put(key, cache.get(key));
+            count++;
+        }
+        cache = temp;
+    }
+
+    @AllArgsConstructor
+    @EqualsAndHashCode
+    private static class CacheKey {
```
navneet1v, line 71: You have provided this very awesome object CacheKey, we should expose this class to other parts of the code so that cache key can be built and used directly for get and put function. 
It will make the interface of get/put easy to use. Now exposing this has downside, lets say going forward if we want to update the cache key, in that case either we can create an internal CacheKey class that can be used, if we don't want all part of code start populating new fields in new cache key. 

Given that thought process I think we should expose this CacheKey class to external parts of the code.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
```diff
@@ -0,0 +1,75 @@
...
+            throw new IllegalArgumentException("ip2geo max cache size must be 0 or greater");
+        }
+        this.cache = CacheBuilder.<CacheKey, Map<String, Object>>builder().setMaximumWeight(maxSize).build();
+    }
+
+    public Map<String, Object> putIfAbsent(
+        final String indexName,
+        final String ip,
+        final Function<String, Map<String, Object>> retrieveFunction
+    ) throws ExecutionException {
```
navneet1v, line 38: we should catch this ExecutionException, and wrap it in another custom Exception and then throw it. We shouldn't expose the internals of Cache to outside parts of the code.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+
+import org.opensearch.common.cache.Cache;
+import org.opensearch.common.cache.CacheBuilder;
```
navneet1v, line 17: why we are going with an internal [OpenSearch cache implementation](https://github.com/opensearch-project/OpenSearch/blob/0ca51a774211184835c4825dfeff38b23198352e/server/src/main/java/org/opensearch/common/cache/Cache.java#L54-L84), when there are other tested and more prominent libraries for cache like Google Guava etc. Example: https://github.com/opensearch-project/k-NN/blob/8e2ad4595bd4f7c64d0b312cac30de8a42e4bf0b/src/main/java/org/opensearch/knn/index/memory/NativeMemoryCacheManager.java#L15-L14

Can we do some deep-dive here to figure out which one to choose and why?

heemin32: Current GeoIP processor use OpenSearch cache implementation which is well tested. Why do we want to explore new option here when there is working option already by adding additional dependency on third party library?

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
heemin32: That is not true singleton; having single instance for each size. Also, there is no use case where we want single instance for each cache size.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
heemin32: `ExecutionException` is from `java.util.concurrent` I don't see much benefit of wrapping it into another exception.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Create a new cache with give size and replace existing cache
+     *
+     * Try to populate the existing value from previous cache to the new cache in best effort
+     *
+     * @param maxSize
+     */
+    public void updateMaxSize(final long maxSize) {
```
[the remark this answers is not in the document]
heemin32: 1. A user can change cache size dynamically through configuration change.
2. There is no public update/delete methods on cached value. Deletion can happen by automatic eviction by policy(cache size in this case). However, it is okay to keep the supposed to be deleted data a little longer.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
heemin32: This is not to throw away all the cached value when size is changed.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
heemin32: I don't no. There is no need for other class to know about it. There is no use case where caller need to construct the key itself. Actually, the interface of get is simpler by hiding the key.

src/main/java/org/opensearch/geospatial/ip2geo/cache/GeoDataCache.java
heemin32: Other option is making this as a private inner class of another class in which this cache is used that I can make in a next PR.

VijayanB (approved):

PR #343, opened 2023-06-26
Add cache layer to reduce GeoIp data retrieval latency

heemin32:
### Description
Add cache layer to reduce GeoIp data retrieval latency
 
### Issues Resolved
N/A
 
### Check List


src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoSettings.java
```diff
@@ -52,16 +52,11 @@ public class Ip2GeoSettings {
...
-     * When the value is 0, it will use default value which will be decided
-     * based on node count and search thread pool size.
+     * Max size for geo data cache
      */
-    public static final Setting<Integer> MAX_CONCURRENT_SEARCHES = Setting.intSetting(
-        "plugins.geospatial.ip2geo.processor.max_concurrent_searches",
-        0,
+    public static final Setting<Long> CACHE_SIZE = Setting.longSetting(
+        "plugins.geospatial.ip2geo.processor.cache_size",
+        1000,
```
vamshin, line 59: why not 10k?

src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
```diff
@@ -37,52 +40,70 @@
...
-    private Map<String, DatasourceMetadata> data;
+    private final GeoIpDataDao geoIpDataDao;
+    private final GeoDataCache geoDataCache;
+    private Map<String, DatasourceMetadata> metadata;
 
-    public Ip2GeoCachedDao(final DatasourceDao datasourceDao) {
+    public Ip2GeoCachedDao(final ClusterService clusterService, final DatasourceDao datasourceDao, final GeoIpDataDao geoIpDataDao) {
         this.datasourceDao = datasourceDao;
+        this.geoIpDataDao = geoIpDataDao;
+        this.geoDataCache = new GeoDataCache(clusterService.getClusterSettings().get(Ip2GeoSettings.CACHE_SIZE));
```
vamshin, line 50: How do we make sure there is only one instance of GeoDataCache?

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoSettings.java
heemin32: This cache is created when Geospatial plugin is installed regardless a user uses Ip2Geo processor or not. So, I think it could be better to start with small number and let user to decide to increase the cache size later. Also, the performance test result with 1,000 wasn't really bad compared to 10,000.

src/main/java/org/opensearch/geospatial/ip2geo/dao/GeoIpDataDao.java
```diff
@@ -242,94 +240,26 @@ public XContentBuilder createDocument(final String[] fields, final String[] valu
...
-                            actionListener.onFailure(e);
-                        }
-                    }
-
-                    @Override
-                    public void onFailure(final Exception e) {
-                        actionListener.onFailure(e);
-                    }
-                })
+                .get(TimeValue.timeValueSeconds(30))
```
martin-gaievski, line 253: please make timeout value a constant

src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
martin-gaievski: maybe add a code comment explaining this reasoning

src/test/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequestTests.java
```diff
@@ -27,4 +29,16 @@ public void testStreamInOut_whenValidInput_thenSucceed() {
         // Verify
         assertEquals(request.getName(), copiedRequest.getName());
     }
+
+    public void testValidate_whenNullOrBlank_thenError() {
+        String datasourceName = Randomness.get().nextInt() % 2 == 0 ? null : " ";
```
martin-gaievski, line 34: I've seen in core such randomized test cases they do in multiple runs, so chance is high that during one run variable will take all possible values. Like a simple for loop with 5 to 10 iterations.

src/main/java/org/opensearch/geospatial/ip2geo/dao/GeoIpDataDao.java
heemin32: Good catch. Missed that.

martin-gaievski (approved):

vamshin (approved):
LGTM! Thanks

PR #370, opened 2023-07-27
Added input parameter validation with few small changes

heemin32:
### Description
1. Added datasource name validation in delete/update api
2. Added datasource name validation in processor creation method
3. Changed index name delimiter from `.` to `-` to be consistent with others
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequest.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequest.java
@@ -15,6 +15,7 @@ import org.opensearch.action.ActionRequest;
 import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.geospatial.ip2geo.common.ParameterValidator;
 
 /**
  * GeoIP datasource delete request
@@ -23,6 +24,7 @@ import org.opensearch.core.common.io.stream.StreamOutput;
 @Setter
 @AllArgsConstructor
 public class DeleteDatasourceRequest extends ActionRequest {
+    private static final ParameterValidator VALIDATOR = new ParameterValidator();
     /**
      * @param name the datasource name
      * @return the datasource name
@@ -43,9 +45,9 @@ public class DeleteDatasourceRequest extends ActionRequest {
     @Override
     public ActionRequestValidationException validate() {
         ActionRequestValidationException errors = null;
-        if (name == null || name.isBlank()) {
+        if (VALIDATOR.validateDatasourceName(name).isEmpty() == false) {
             errors = new ActionRequestValidationException();
-            errors.addValidationError("Datasource name should not be empty");
+            errors.addValidationError("no such datasource exist");
         }
         return errors;
     }
--- a/src/test/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidatorTests.java
+++ b/src/test/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidatorTests.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.geospatial.ip2geo.common;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import org.opensearch.common.Randomness;
+import org.opensearch.common.Strings;
+import org.opensearch.geospatial.GeospatialTestHelper;
+import org.opensearch.geospatial.ip2geo.Ip2GeoTestCase;
+
+public class InputFormatValidatorTests extends Ip2GeoTestCase {
+    public void testValidateDatasourceName_whenValidName_thenSucceed() {
+        ParameterValidator inputFormatValidator = new ParameterValidator();
+        String validDatasourceName = GeospatialTestHelper.randomLowerCaseString();
+
+        // Run
+        List<String> errorMsgs = inputFormatValidator.validateDatasourceName(validDatasourceName);
+
+        // Verify
+        assertTrue(errorMsgs.isEmpty());
+    }
+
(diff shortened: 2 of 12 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        if (datasourceName.contains("#")) {
+            return "datasource name must not contain '#'";
+        }
+        if (datasourceName.contains(":")) {
+            return "datasource name must not contain ':'";
+        }
+        if (datasourceName.charAt(0) == '_' || datasourceName.charAt(0) == '-' || datasourceName.charAt(0) == '+') {
+            return "datasource name must not start with '_', '-', or '+'";
+        }
```
junqiu-lei, line 31: Looks like it can validate only one item each time and return, if there have multiple errors at one time, can we point out them together?

heemin32: W can. Let me try to update.

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequest.java
```diff
@@ -43,9 +45,9 @@ public DeleteDatasourceRequest(final StreamInput in) throws IOException {
     @Override
     public ActionRequestValidationException validate() {
         ActionRequestValidationException errors = null;
-        if (name == null || name.isBlank()) {
+        if (VALIDATOR.validateDatasourceName(name) != null) {
```
martin-gaievski, line 48: nit: you can use Objects.nonNull()

src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.UnsupportedEncodingException;
+
+import org.opensearch.OpenSearchException;
+import org.opensearch.common.Strings;
+
+public class InputFormatValidator {
+    private static final int MAX_DATASOURCE_NAME_BYTES = 127;
+
+    public String validateDatasourceName(final String datasourceName) {
+        if (datasourceName == null || datasourceName.isEmpty()) {
```
martin-gaievski, line 17: There should be some method in StringUtils for checking exactly this

src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
```diff
@@ -0,0 +1,47 @@
...
+            return "datasource name must not be empty";
+        }
+        if (!Strings.validFileName(datasourceName)) {
+            return "datasource name must not contain the following characters " + Strings.INVALID_FILENAME_CHARS;
```
martin-gaievski, line 21: should we use String.format or join instead of direct concatination

src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        if (datasourceName.contains(":")) {
+            return "datasource name must not contain ':'";
+        }
+        if (datasourceName.charAt(0) == '_' || datasourceName.charAt(0) == '-' || datasourceName.charAt(0) == '+') {
+            return "datasource name must not start with '_', '-', or '+'";
+        }
+        int byteCount = 0;
+        try {
+            byteCount = datasourceName.getBytes("UTF-8").length;
```
martin-gaievski, line 34: we can use StandardCharsets.UTF_8 instead of using own string constant

src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.opensearch.geospatial.ip2geo.common;
+
+import java.io.UnsupportedEncodingException;
+
+import org.opensearch.OpenSearchException;
+import org.opensearch.common.Strings;
+
+public class InputFormatValidator {
```
martin-gaievski, line 13: Please add description for this public class

src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
heemin32: ack

src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
heemin32: ack

src/main/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidator.java
heemin32: ack

src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceRequest.java
heemin32: Thanks. I feel `!= null` is more readable. Anyway, this code will be removed with update.

junqiu-lei (approved):

src/test/java/org/opensearch/geospatial/ip2geo/common/InputFormatValidatorTests.java
```diff
@@ -0,0 +1,72 @@
...
+            longName.toString(),
+            "name is too long",
+            point,
+            "must not be '.' or '..'"
+        );
+
+        for (Map.Entry<String, String> entry : nameToError.entrySet()) {
+
+            // Run
+            System.out.println(entry.getKey());
```
martin-gaievski, line 64: do we need this line?

heemin32: Oops. Added by mistake.

martin-gaievski (approved):

PR #379, opened 2023-08-07
[Refactor] Strings utility methods to core library

heemin32:
### Description
[Refactor] Strings utility methods to core library
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/action/RestGetDatasourceHandler.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/action/RestGetDatasourceHandler.java
@@ -12,7 +12,7 @@ import static org.opensearch.rest.RestRequest.Method.GET;
 import java.util.List;
 
 import org.opensearch.client.node.NodeClient;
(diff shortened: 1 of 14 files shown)


navneet1v (approved):

src/main/java/org/opensearch/geospatial/ip2geo/action/RestGetDatasourceHandler.java
```diff
@@ -12,7 +12,7 @@
 import java.util.List;
 
 import org.opensearch.client.node.NodeClient;
-import org.opensearch.common.Strings;
+import org.opensearch.core.common.Strings;
```
martin-gaievski, line 15: We don't need Strings class, unless it's used somewhere else in our plugin we need to remove this import

heemin32: It is used in this class.

martin-gaievski (approved):

PR #382, opened 2023-08-15
Revert datasource state when delete fails

heemin32:
### Description
Revert datasource state when delete fails
 
### Issues Resolved
N/A
 
### Check List

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
@@ -111,9 +111,18 @@ public class DeleteDatasourceTransportAction extends HandledTransportAction<Dele
         if (datasource == null) {
             throw new ResourceNotFoundException("no such datasource exist");
         }
-
+        DatasourceState previousState = datasource.getState();
         setDatasourceStateAsDeleting(datasource);
-        geoIpDataDao.deleteIp2GeoDataIndex(datasource.getIndices());
+
+        try {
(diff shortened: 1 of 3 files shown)


src/main/java/org/opensearch/geospatial/ip2geo/action/DeleteDatasourceTransportAction.java
```diff
@@ -111,9 +111,18 @@ protected void deleteDatasource(final String datasourceName) throws IOException
...
         }
-
+        DatasourceState previousState = datasource.getState();
         setDatasourceStateAsDeleting(datasource);
-        geoIpDataDao.deleteIp2GeoDataIndex(datasource.getIndices());
+
+        try {
+            geoIpDataDao.deleteIp2GeoDataIndex(datasource.getIndices());
+        } catch (Exception e) {
+            if (previousState.equals(datasource.getState()) == false) {
```
martin-gaievski, line 120: what is the scenario when datasource has previous state? We're setting it to "deleting" before we call delete method in setDatasourceStateAsDeleting().

heemin32: We first change the datasource state to "deleting" and run the delete process. So, datasource always has previous state either "available", "create_failed", or "deleting".

martin-gaievski: Got more info on scenario, we need to delete two resources(index and metadata), if we fail to delete metadata, we don’t revert state because index for data is already deleted

martin-gaievski (approved):

junqiu-lei (approved):

PR #677, opened 2024-08-27
Use instance of LockService instantiated in JobScheduler through Guice

cwperks:
### Description

Companion JS PR: https://github.com/opensearch-project/job-scheduler/pull/670

This PR shows how Geospatial can be refactored to use the instance of the LockService that is initialized in Job Scheduler's `createComponents`.

This PR is part of an effort to remove usages of `ThreadContext.stashContext` across the plugins: https://github.com/opensearch-project/opensearch-plugins/issues/238

Currently, geospatial instantiates its own instance of the LockService by passing in the Client given to geospatial through `createComponents`.

As part of the effort to [Strengthen System Indices in the Plugin Ecosystem](https://github.com/opensearch-project/security/issues/4439), plugins will be restricted to only perform transport actions to their own system indices. This PR is to ensure that Geospatial uses the LockService instantiated by JS (which has permission to JS system indices) vs creating its own LockService. 

### Related Issues

Related to https://github.com/opensearch-project/security/issues/4439

### Check List
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
@@ -67,6 +67,9 @@ public class Ip2GeoLockService {
      * @param listener the listener
      */
     public void acquireLock(final String datasourceName, final Long lockDurationSeconds, final ActionListener<LockModel> listener) {
+        if (lockService == null) {
+            throw new OpenSearchException("Ip2GeoLockService is not initialized");
+        }
         lockService.acquireLockWithId(JOB_INDEX_NAME, lockDurationSeconds, datasourceName, listener);
     }
 
@@ -78,6 +81,9 @@ public class Ip2GeoLockService {
      * @return lock model
      */
     public Optional<LockModel> acquireLock(final String datasourceName, final Long lockDurationSeconds) {
+        if (lockService == null) {
+            throw new OpenSearchException("Ip2GeoLockService is not initialized");
+        }
         AtomicReference<LockModel> lockReference = new AtomicReference();
         CountDownLatch countDownLatch = new CountDownLatch(1);
         lockService.acquireLockWithId(JOB_INDEX_NAME, lockDurationSeconds, datasourceName, new ActionListener<>() {
@@ -108,6 +114,9 @@ public class Ip2GeoLockService {
      * @param lockModel the lock model
      */
     public void releaseLock(final LockModel lockModel) {
+        if (lockService == null) {
+            throw new OpenSearchException("Ip2GeoLockService is not initialized");
+        }
         lockService.release(
             lockModel,
             ActionListener.wrap(released -> {}, exception -> log.error("Failed to release the lock", exception))
@@ -121,6 +130,9 @@ public class Ip2GeoLockService {
      * @return renewed lock if renew succeed and null otherwise
      */
     public LockModel renewLock(final LockModel lockModel) {
+        if (lockService == null) {
+            throw new OpenSearchException("Ip2GeoLockService is not initialized");
+        }
         AtomicReference<LockModel> lockReference = new AtomicReference();
         CountDownLatch countDownLatch = new CountDownLatch(1);
         lockService.renewLock(lockModel, new ActionListener<>() {
--- a/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
@@ -14,7 +14,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.opensearch.OpenSearchException;
-import org.opensearch.client.Client;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.jobscheduler.spi.LockModel;
@@ -31,17 +30,6 @@ public class Ip2GeoLockService {
     public static final long RENEW_AFTER_IN_SECONDS = 120l;
     private final ClusterService clusterService;
     private LockService lockService;
-
-    /**
-     * Constructor
-     *
-     * @param clusterService the cluster service
-     * @param client the client
-     */
-    public Ip2GeoLockService(final ClusterService clusterService, final Client client) {
-        this.clusterService = clusterService;
-        this.lockService = new LockService(client, clusterService);
-    }
 
     /**
      * Constructor
--- a/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
@@ -30,7 +30,7 @@ public class Ip2GeoLockService {
     public static final long LOCK_DURATION_IN_SECONDS = 300l;
     public static final long RENEW_AFTER_IN_SECONDS = 120l;
     private final ClusterService clusterService;
-    private final LockService lockService;
+    private LockService lockService;
 
     /**
      * Constructor
@@ -41,6 +41,19 @@ public class Ip2GeoLockService {
     public Ip2GeoLockService(final ClusterService clusterService, final Client client) {
         this.clusterService = clusterService;
         this.lockService = new LockService(client, clusterService);
+    }
+
+    /**
+     * Constructor
+     *
+     * @param clusterService the cluster service
+     */
+    public Ip2GeoLockService(final ClusterService clusterService) {
+        this.clusterService = clusterService;
+    }
+
+    public void initialize(final LockService lockService) {
+        this.lockService = lockService;
     }
 
     /**
--- a/src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
+++ b/src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
@@ -9,17 +9,24 @@ import static org.opensearch.geospatial.ip2geo.jobscheduler.Datasource.IP2GEO_DA
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Supplier;
 
 import org.opensearch.action.ActionRequest;
+import org.opensearch.action.search.PitService;
 import org.opensearch.client.Client;
 import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
+import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.node.DiscoveryNodes;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.collect.MapBuilder;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.common.lifecycle.Lifecycle;
 import org.opensearch.common.lifecycle.LifecycleComponent;
+import org.opensearch.common.lifecycle.LifecycleListener;
 import org.opensearch.common.settings.ClusterSettings;
 import org.opensearch.common.settings.IndexScopedSettings;
 import org.opensearch.common.settings.Setting;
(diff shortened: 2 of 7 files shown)


cwperks:
The CI will fail until companion JS PR is merged

heemin32:
@cwperks Could you also update the change log?

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
```diff
@@ -265,4 +275,52 @@ public List<AggregationSpec> getAggregations() {
 
         return List.of(geoHexGridSpec);
     }
+
+    @Override
+    public void onNodeStarted(DiscoveryNode localNode) {
+        LockService lockService = GuiceHolder.getLockService();
+        ip2GeoLockService.initialize(lockService);
+
+        /**
```
heemin32, line 284: We can remove this comment now

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
```diff
@@ -43,6 +43,19 @@ public Ip2GeoLockService(final ClusterService clusterService, final Client clien
         this.lockService = new LockService(client, clusterService);
```
heemin32, line 43: Shouldn't we block this constructor if we don't want plugin to instantiate their own lock service?

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
```diff
@@ -43,6 +43,19 @@ public Ip2GeoLockService(final ClusterService clusterService, final Client clien
...
     }
 
+    /**
+     * Constructor
+     *
+     * @param clusterService the cluster service
+     */
+    public Ip2GeoLockService(final ClusterService clusterService) {
+        this.clusterService = clusterService;
```
heemin32, line 52: We might need to add lockService not initialized check in every Ip2GeoLockService methods.

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
```diff
@@ -30,7 +30,7 @@ public class Ip2GeoLockService {
     public static final long LOCK_DURATION_IN_SECONDS = 300l;
     public static final long RENEW_AFTER_IN_SECONDS = 120l;
     private final ClusterService clusterService;
-    private final LockService lockService;
+    private LockService lockService;
```
heemin32, line 33: Wouldn't it be better to make Ip2GeoLockService to be LifecycleComponent and get injected both cluster service and lock service?

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
cwperks: Removed this comment

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
cwperks: Yes, but first all usages in the default distribution need to be addressed: https://github.com/search?q=org%3Aopensearch-project%20LockService&type=code

I'll raise a PR in JS to formally mark the constructor as Deprecated

heemin32: Oh. I mean, `public Ip2GeoLockService(final ClusterService clusterService, final Client client)` not `public LockService(final ClusterService clusterService, final Client client)`

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
cwperks: It is still being returned from `createComponents` so that its injectable. The difference in this PR is that it now implements `onNodeStarted` to get the instance of LockService initialized by the job scheduler from the GuiceHolder and then sets it on the Ip2GeoLockService (Lock Service Wrapper) instance.

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
cwperks: yes, good point. The tests need to be updated in Ip2GeoLockServiceTests. Taking a look now.

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
cwperks: It should never happen (i.e. onNodeStarted will always be called and the lock service will get set accordingly) but these checks would be good to put in none-the-less.

cwperks: Added the checks

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
```diff
@@ -265,4 +275,48 @@ public List<AggregationSpec> getAggregations() {
...
+    @Override
+    public void onNodeStarted(DiscoveryNode localNode) {
+        LockService lockService = GuiceHolder.getLockService();
+        ip2GeoLockService.initialize(lockService);
+
+        DatasourceRunner.getJobRunnerInstance()
+            .initialize(this.clusterService, this.datasourceUpdateService, this.ip2GeoExecutor, this.datasourceDao, this.ip2GeoLockService);
+    }
+
+    public static class GuiceHolder implements LifecycleComponent {
```
heemin32, line 288: Can we move this class to its own file?

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
```diff
@@ -265,4 +275,48 @@ public List<AggregationSpec> getAggregations() {
...
+
+        private static LockService lockService;
+
+        @Inject
+        public GuiceHolder(final LockService lockService) {
+            GuiceHolder.lockService = lockService;
+        }
+
+        public static LockService getLockService() {
```
heemin32, line 297: Might be good to make it as package private scope so that this method is not being used widely in the code.
This static pattern should be restricted to minimize its impact as it make hard to write a unit test.
```suggestion
        static LockService getLockService() {
```

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
heemin32: I was thinking like
```
public class Ip2GeoLockService implements LifecycleComponent { 
  private final static Ip2GeoLockService lockService;
  
  public static Ip2GeoLockService getLockService() {
    return lockService;
  }

  @Inject
  public Ip2GeoLockService(final ClusterService clusterService, final LockService lockService) {
  }
}


@Override
public Collection<Class<? extends LifecycleComponent>> getGuiceServiceClasses() {
    final List<Class<? extends LifecycleComponent>> services = new ArrayList<>(2);
    services.add(Ip2GeoListener.class);
    services.add(Ip2GeoLockService.class);
    return services;
}
```
This way, we don't need an additional class.  
However, this approach opens up the possibility for the `public static Ip2GeoLockService getLockService()` method to be used across different parts of the code. Therefore, I think having a separate class might be better in that regard.

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
cwperks: Removed the constructor of `Ip2GeoLockService` that accepts a `Client`. There are still instances where `LockService` is instantiated in tests that I plan to remove in a subsequent PR. The difficulty I am facing now is creating a mock instance of LockService because its a final class in the JS SPI.

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
cwperks: Its a static inner class so I think it needs to be defined in this file. This is following a similar pattern present in the [security plugin](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L2194-L2262). This is a way to do dependency injection outside of the transport layer.

heemin32: It does not matter where and how the class is defined. As long as you return your class in `getGuiceServiceClasses()`, the dependency will get injected. 

`    @Override
    public Collection<Class<? extends LifecycleComponent>> getGuiceServiceClasses() {
        final List<Class<? extends LifecycleComponent>> services = new ArrayList<>(2);
        services.add(Ip2GeoListener.class);
        services.add(GuiceHolder.class);
        return services;
    }
`
Anyway, I think having this class inside Plugin class might make sense if we want it to be used only inside this Plugin class.
private or package private might be better instead of public then.

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
cwperks: That looks like a good solution, but how do you refer to the instance of the LifecycleComponent that's created?

For instance this line:

```
DatasourceRunner.getJobRunnerInstance()
            .initialize(this.clusterService, this.datasourceUpdateService, this.ip2GeoExecutor, this.datasourceDao, this.ip2GeoLockService);
```

Takes the `ip2GeoLockService` as the last arg. When using the service class, how do you get a hold of an instance?

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
cwperks: Switched to package-private

cwperks: Looks like it needs to be public:

ref: https://github.com/opensearch-project/geospatial/actions/runs/12895082709/job/35955219215?pr=677

```
?  java.lang.AssertionError: Wrong access modifiers on public org.opensearch.geospatial.plugin.GeospatialPlugin$GuiceHolder(org.opensearch.jobscheduler.spi.utils.LockService)
?  	at org.opensearch.common.inject.DefaultConstructionProxyFactory$1.newInstance(DefaultConstructionProxyFactory.java:69)
?  	at org.opensearch.common.inject.ConstructorInjector.construct(ConstructorInjector.java:101)
?  	at org.opensearch.common.inject.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:140)
?  	at org.opensearch.common.inject.InjectorImpl$4$1.call(InjectorImpl.java:753)
?  	at org.opensearch.common.inject.InjectorImpl.callInContext(InjectorImpl.java:809)
?  	at org.opensearch.common.inject.InjectorImpl$4.get(InjectorImpl.java:748)
?  	at org.opensearch.common.inject.InjectorImpl.getInstance(InjectorImpl.java:[792](https://github.com/opensearch-project/geospatial/actions/runs/12895082709/job/35955219215?pr=677#step:6:793))
?  	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
?  	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708)
?  	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
?  	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
?  	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
?  	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
?  	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
?  	at org.opensearch.node.Node.<init>(Node.java:1548)
?  	at org.opensearch.node.Node.<init>(Node.java:452)
```

src/main/java/org/opensearch/geospatial/ip2geo/common/Ip2GeoLockService.java
heemin32: You can call `Ip2GeoLockService.getLockService()`

However, I think having a `GuiceHolder` class might be better not to expose that static method.

src/main/java/org/opensearch/geospatial/plugin/GeospatialPlugin.java
heemin32: In such case, making methods as package private might be enough.

cwperks: Any of the overridden methods need to be marked public. I marked `getLockService` as package-private.

heemin32: Yes. I was talking only for those static methods :)

heemin32 (approved):
LGTM. Thanks!

junqiu-lei (approved):

Issue #698, opened 2024-11-18
(Draft) [RFC] Expose Geo2IP Enrichment feature to other plugins to standardize OpenSearch IP enrichment offering


andy-k-improving:
The purpose of this RFC (request for comments) is to gather community feedbacks on a proposal to provide a way to expose the IP enrichment functionality as a registered Action on OpenSearch, which will allow other OpenSearch Plugins to leverage the enrichment, instead of reinventing the wheel.

# Problem Statement
At the moment the `GeoSpatial` plugin do offer IP enrichment feature, however the scope is limited to Document ingestion and also within the `GeoSpatial` plugin itself.
This segregation will force other Plugin within the Eco system to implement their own IP enrichment feature, even when `GeoSpatial` plugin is present in the system.
And this result in fragmentation and duplicate effort on the IP enrichment capability for OpenSearch platform as a whole. 

# Current State
TBC

# Proposal
TBC

# Approach
TBC

# Data flow Diagram
TBC

# API Design
TBC

heemin32:
Hi @andy-k-improving, it's good to know that other plugins might need to use the geo2ip feature. For a plugin to leverage the geo2ip functionality, I believe it would require a datasource as an input parameter.

heemin32:
Feel free to proceed with publishing the PR, and we can continue the discussion there.

andy-k-improving:
> Feel free to proceed with publishing the PR, and we can continue the discussion there.
> 
> Based on [opensearch-project/sql#3038](https://github.com/opensearch-project/sql/issues/3038), it seems it might be useful to have optional field name as parameter input as well.

That make sense, will update the RFC accordingly.
Cool, then I will proceed, thanks!

andy-k-improving:
Implementation available under:
https://github.com/opensearch-project/geospatial/pull/700

dblock:
[[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/dblock), [2](https://github.com/reta), [3](https://github.com/andrross/), [4](https://github.com/krisfreedain)]

PR #700, opened 2024-11-23
[Feature] New module `geospatial-client` for cross-plugin IP enrichment 

andy-k-improving:
### Description
This is an implementation ticket which follow the discussion of https://github.com/opensearch-project/geospatial/issues/698, to create a new Gradle module for other OpenSearch Plugin to import, in order to make an IP enrichment call to GeoSpatial plugin to resolve GeoLocation data with existing data source. 

Hight-level code changes:
 - A new Gradle module `geospatial-client` with appreciate Gradle setting files to publish it as a standalone Jar. 
 - A core class `ipEnrichmentActionClient` which serve as the interface, this is the only way for other Plugin to interact with GeoSpatial's IP enrichment functionality
 - A collection of classes to support the IP enrichment remote call access `IpEnrichmentAction`, `IpEnrichmentRequest` and `IpEnrichmentResponse`
 - `IpEnrichementTransportAction` class which contain the logic for the IP address lookup and register it during the GeoSpatial's bootstrap time

Non GeoSpatial plugin can now interact with the new functionality like below:
```
IpEnrichmentActionClient ipClient = new IpEnrichmentActionClient(client.getNodeClient());
        try {
          Map<String, Object> geoLocationData = ipClient.getGeoLocationData("50.68.18.229");
        } catch (Exception e) {
            throw new RuntimeException(e);
        }
```

### Related Issues
Resolves https://github.com/opensearch-project/geospatial/issues/698

### Check List
- [x] New functionality includes testing.
- [x] Commits are signed per the DCO using `--signoff`.


andy-k-improving:
@heemin32 here is the implementation for https://github.com/opensearch-project/geospatial/issues/698 as we discussed earlier of the week, would you mind to have a look and advise?
Any comment will be appreciated.

client/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+apply plugin: 'java'
+apply plugin: 'maven-publish'
+apply plugin: 'jacoco'
+apply plugin: 'io.freefair.lombok'
+
+allprojects {
+    group = opensearch_group
+    version = "${opensearch_build}"
+    targetCompatibility = JavaVersion.VERSION_11
+    sourceCompatibility = JavaVersion.VERSION_11
```
navneet1v, line 15: is this compatibility correct? I thought whole opensearch moved to java 17.

client/src/test/java/org/opensearch/geospatial/action/IpEnrichmentActionClientTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class IpEnrichmentActionClientTest {
```
navneet1v, line 26: For all the tests lets use the same test framework used in plugin rather than using standard Mockito ensure that test framework is same across different modules of the plugin.

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentActionClient.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Facade to provide GeoLocation enrichment for other plugins.
+ */
+@Log4j2
+@AllArgsConstructor
+public class IpEnrichmentActionClient {
+
+    NodeClient nodeClient;
```
navneet1v, line 25: any reason for NodeClient being package private?

client/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+}
+
+repositories {
+    mavenLocal()
+    maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
+    mavenCentral()
+    maven { url "https://plugins.gradle.org/m2/" }
+}
+
+compileJava {
```
navneet1v, line 25: can we add spotlessJava task or it is already imported from root module?

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentAction.java
```diff
@@ -0,0 +1,24 @@
...
+package org.opensearch.geospatial.action;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.core.action.ActionResponse;
+
+/**
+ * An ActionType registered on OpenSearch registry, for inter-cluster transportAction call,
+ * to resolve GeoLocation for IP String.
+ */
+public class IpEnrichmentAction extends ActionType<ActionResponse> {
```
navneet1v, line 15: not sure why this action is in client? and its transport action is in root module. Please correct me if I am wrong here. 
One client implementation in Opensearch project I have seen is this: https://github.com/opensearch-project/ml-commons/tree/main/client .

navneet1v (commented):
Looked very high level on the code. I like the idea of having a client. But I want to get some consistencies in the code. and I am hoping client to very lightweight.

client/build.gradle
andy-k-improving: Actually I think that make more sense to enforce the same from `opensearch.build` instead, in this way the Java version requirement of this module will always identical to OpenSearch main repo.

andy-k-improving: I replaced above with `apply plugin: 'opensearch.build'` instead.

client/src/test/java/org/opensearch/geospatial/action/IpEnrichmentActionClientTest.java
andy-k-improving: I have updated the build.gradle to read the Mockito version from the `opensearch.build` plugin, in this case we will have the same version of Mockito as the main OpenSearch repo.
```
testImplementation "org.mockito:mockito-core:${versions.mockito}"
```
Or you mean I should ditch Mockito all together?

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentActionClient.java
andy-k-improving: Nope, updated with `final private` :)

client/build.gradle
andy-k-improving: Done, by adding it on the build.gralde explicitl.

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentAction.java
andy-k-improving: This is needed because during the plugin runtime, other plugin need to submit the `ActionType` as an identifier for one of the argument on `nodeClient.execute(ActionType, xxx, xx)`.
I have also checked on the above ML-common integration, indeed they also expose the respective action also `MLPredictionTaskAction`, but they decided on a separate module call `opensearch-ml-common` instead. 

I think both approach work, either we can keep everything in a jar and only expose `geoSpatial-client`, or we can follow the ML plugin approach, which expose TWO jars `ml-common` and `ml-client`, but essential contains the same classes.

andy-k-improving:
@navneet1v I have fixed most of the point except the Mockito, can you have a look a review?
Thanks,

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentActionClient.java
```diff
@@ -0,0 +1,57 @@
...
+            IpEnrichmentAction.INSTANCE,
+            new IpEnrichmentRequest(ipString, datasourceName)
+        );
+        // Send out the request and process the response.
+        try {
+            ActionResponse genericActionResponse = responseActionFuture.get();
+            IpEnrichmentResponse enrichmentResponse = IpEnrichmentResponse.fromActionResponse(genericActionResponse);
+            return enrichmentResponse.getGeoLocationData();
+        } catch (Exception e) {
+            log.error("GeoSpatial IP Enrichment call failure, with detail: ", e);
```
heemin32, line 53: I believe we should throw an exception and let the caller handle it, rather than returning `null`?

src/main/java/org/opensearch/geospatial/ip2geo/action/IpEnrichmentTransportAction.java
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+            String dataSourceName = Optional.ofNullable(enrichmentRequest.getDatasourceName()).orElse(defaultDataSourceName);
+            String indexName = ip2GeoCachedDao.getIndexName(dataSourceName);
+            Map<String, Object> geoLocationData = ip2GeoCachedDao.getGeoData(indexName, ipString);
+            log.debug("GeoSpatial IP lookup on IP: [{}], and result [{}]", ipString, geoLocationData);
+            listener.onResponse(new IpEnrichmentResponse(geoLocationData));
+        }
+    }
+
+    private String getDefaultDataSourceName() {
```
heemin32, line 82: Would it be better to make the data source name mandatory? Allowing the caller to omit the data source and selecting one randomly from the available options might not be ideal.

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentResponse.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * Static method to convert a given ActionResponse to IpEnrichmentResponse by serialisation with streamOuput.
+     * This will be required for cross plugin communication scenario, as multiple class definition will be loaded
+     * by respective Plugin's classloader.
+     * @param actionResponse An IpEnrichmentResponse in casted-up form.
+     * @return An IpEnrichmentResponse object which contain the same payload as the incoming object.
+     */
+    public static IpEnrichmentResponse fromActionResponse(ActionResponse actionResponse) {
```
heemin32, line 66: How would it work if we need to modify either the request or the response? Will these changes be handled correctly, particularly during a cluster upgrade? Additionally, could there be scenarios where other plugins use a different version of the client compared to the geospatial plugin?

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentActionClient.java
andy-k-improving: Done, updating it to throw exception instead.

andy-k-improving:
@heemin32 I updated the code to reflect comments, can you have another look.
Also whats your thought on the cluster upgrade scenario?

heemin32:
> @heemin32 I updated the code to reflect comments, can you have another look. Also whats your thought on the cluster upgrade scenario?

I think, we can use add versioning on this serialization/deserialization using OpenSearch version.

heemin32:
@Xtansia Could you take a look at this?

andy-k-improving:
> I think, we can add versioning on this serialization/deserialization using OpenSearch vers

Just to align on this, what's the expected behaviour in the case when  serializationID doesn't match?
Should we thrown an exception, we attempt to parse it as per the serializationID? (Ex: v1.19.0 version of the plugin serializating the object from v1.18.0...etc)

andy-k-improving:
> > @heemin32 I updated the code to reflect comments, can you have another look. Also whats your thought on the cluster upgrade scenario?
> 
> I think, we can add versioning on this serialization/deserialization using OpenSearch version?

I have updated the code to perform version, can you have a look and advise?
Thx!

heemin32:
Another general comment is if we should implement it as a rest API rather than client? What do you think? @andy-k-improving

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentResponse.java
```diff
@@ -0,0 +1,92 @@
...
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream(); OutputStreamStreamOutput osso = new OutputStreamStreamOutput(baos)) {
+            actionResponse.writeTo(osso);
+
+
+            try (StreamInput input = new InputStreamStreamInput(new ByteArrayInputStream(baos.toByteArray()))) {
+                String objectVersion = input.readString();
+                if (VERSION.equals(objectVersion)) {
+                    return new IpEnrichmentResponse(input);
+                } else {
+                    throw new IllegalArgumentException("Fail to serialise IpEnrichmentResponse due to version mismatch");
```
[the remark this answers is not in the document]
andy-k-improving: Right, let me revert change, but to clarify on the context.
If you mean the backward competiable, we are referring something similar to https://github.com/opensearch-project/OpenSearch/blob/581771053c324e0774c3309ec9806d53d99f9f35/server/src/main/java/org/opensearch/action/admin/indices/upgrade/post/ShardUpgradeResult.java#L49

Which read the `Version` object from the constructor and use it on the serialisation || de-serailisation process to identify the version mis-match and handle the backward compatibility accordingly?

andy-k-improving:
> Another general comment is if we should implement it as a rest API rather than client? What do you think? @andy-k-improving

We should have it in the long term || as day-2, but I don't think that make sense to bundle it in one PR, as there will be more considerations I foresee (Ex: performance, web-layer security) need to be addressed || discussed ahead of the implementation. 
Let me create another issue perhaps?

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentResponse.java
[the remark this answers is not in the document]
heemin32: Yes. For example. https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/index/query/parser/KNNQueryBuilderParser.java#L142-L163

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentRequest.java
```diff
@@ -0,0 +1,103 @@
...
+    private String datasourceName;
+
+    /**
+     * Constructor for TransportAction.
+     * @param streamInput the streamInput.
+     */
+    public IpEnrichmentRequest(StreamInput streamInput) throws IOException {
+        super(streamInput);
+        ipString = streamInput.readString();
+        datasourceName = streamInput.readOptionalString();
```
heemin32, line 45: Let's make it as non optional.

andy-k-improving:
@heemin32 Also do you foresee any problem to publish the `geospatial-client` jar? 
In local, I see the jar is produced correctly, but not sure will that be the same when publishing to AWS || Maven central.

heemin32:
> Let me create another issue perhaps?

No need for that. I was considering either this or that, but not both. Thanks.

heemin32:
> @heemin32 Also do you foresee any problem to publish the `geospatial-client` jar? In local, I see the jar is produced correctly, but not sure will that be the same when publishing to AWS || Maven central.

I don’t have much expertise on that topic. @ylwu-amzn, is there anything specific we should consider when publishing the client? Asking as ml-common already has done it.

client/src/main/java/org/opensearch/geospatial/action/IpEnrichmentRequest.java
andy-k-improving: Done.

andy-k-improving:
> No need for that. I was considering either this or that, but not both. Thanks.

If that is the case, then I rather keep this instead of REST.

heemin32 (approved):

src/test/java/org/opensearch/geospatial/ip2geo/action/IpEnrichmentTransportActionTests.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    public void testDoExecute_All_Succeed() {
+        IpEnrichmentRequest request = new IpEnrichmentRequest("192.168.1.1", "testSource");
+        action.doExecute(task, request, listener);
+
+        verify(listener, times(1)).onResponse(any(IpEnrichmentResponse.class));
+    }
+    
+
+    /**
```
heemin32, line 58: Please remove this test case.

heemin32: Or, rename it.

andy-k-improving: Removed.

andy-k-improving:
Rebased again for the JDK 21 changes on main, waiting for CI.

andy-k-improving:
@navneet1v Hi, I have updated the code and tested the build command locally, it should work by now.
Would you mind to have another look and advise?
Thanks,

andy-k-improving:
Updated the code style should be ok for review again.

src/main/java/org/opensearch/geospatial/ip2geo/action/IpEnrichmentTransportAction.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param task the task.
+     * @param request request object in the form of IpEnrichmentRequest which contain the IP String to resolve
+     * @param listener a container which encapsulate IpEnrichmentResponse object with the GeoLocation data for given IP.
+     */
+    @Override
+    protected void doExecute(Task task, ActionRequest request, ActionListener<ActionResponse> listener) {
+        IpEnrichmentRequest enrichmentRequest = IpEnrichmentRequest.fromActionRequest(request);
+        String ipString = enrichmentRequest.getIpString();
+        if (enrichmentRequest.getDatasourceName() == null) {
+            log.error("No data source available, IpEnrichmentTransportAction aborted.");
```
navneet1v, line 57: is this error due to user configuration or related to system?

andy-k-improving: User configuration, but actually this will never happen, because OpenSearch runtime should run `ipEnrichmentRequest.validate( )` ahead, so when TransportAction received the request object. 
The dataSource String should be present.

src/main/java/org/opensearch/geospatial/ip2geo/action/IpEnrichmentTransportAction.java
```diff
@@ -0,0 +1,67 @@
...
+            listener.onFailure(new IllegalArgumentException());
```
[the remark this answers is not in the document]
andy-k-improving: Same, actually the `if` clause can be removed altogether, hence there is no need for this exception.

andy-k-improving:
Hi @navneet1v, would you mind to have another look?
Thanks,

client/src/test/java/org/opensearch/geospatial/action/IpEnrichmentActionClientTest.java
heemin32: It seems we don't extend from `OpenSearchTestCase` for client module? https://github.com/opensearch-project/ml-commons/blob/main/client/src/test/java/org/opensearch/ml/client/MachineLearningNodeClientTest.java

heemin32:
Could you rebase and see if the failing CI will succeed?

andy-k-improving:
@heemin32 done.

heemin32:
@andy-k-improving, thanks for the change. If you don't get response from @navneet1v by next Monday, let me pull in another person to review this PR.

client/src/test/java/org/opensearch/geospatial/action/IpEnrichmentRequestTests.java
```diff
@@ -0,0 +1,71 @@
...
+ * Test cases for IpEnrichmentRequest.
+ */
+public class IpEnrichmentRequestTests {
+
+    /**
+     * Test validate() against a valid record.
+     */
+    @Test
+    public void testValidateValidRequest() {
+        System.out.println("Test");
```
navneet1v, line 21: remove this.

client/src/test/java/org/opensearch/geospatial/action/IpEnrichmentActionClientTest.java
navneet1v: The problem with not doing this would be that we first we will not able to test the code with random data and another problem is since we will not have seed whenever we want to reproduce a failure. Not sure why MLCommons didn't do this. This is an acceptable practice in Opensearch Plugins. Plus you will miss out on all the different mocking and testing framework added by Opensearch.

navneet1v (commented):
Small comment. Overall code looks good to me. There is system.out statement left in the code. Added a comment for that.

client/src/test/java/org/opensearch/geospatial/action/IpEnrichmentActionClientTest.java
andy-k-improving: In my point, this depends on the amount of logic to test. 
I agree we should have it to remain the consistency, but for now if we aim to have it as a thin client which has no actual logic inside except the argument validation.
This may be a overkill.

client/src/test/java/org/opensearch/geospatial/action/IpEnrichmentRequestTests.java
andy-k-improving: Done.

navneet1v (approved):

PR #715, opened 2025-01-28
Replace usages of ThreadContext.stashContext with pluginSubject.runAs

cwperks:
### Description

This PR replaces usages of ThreadContext.stashContext with a replacement that enforces stricter ownership over system indices. Plugins can use this replacement for system index access and the advantage of this replacement is that it provides context into which plugin is performing privileged actions like system index access.

### Related Issues

Resolves https://github.com/opensearch-project/opensearch-plugins/issues/238

### Check List
- [ ] Commits are signed per the DCO using `--signoff`.


cwperks:
> Hi @cwperks , should this change be included in 2.19 release? The code freeze for 2.19 starts at Jan 28th as per: https://opensearch.org/releases.html

No this does not need to be included. This is part of a larger effort across [plugins](https://github.com/opensearch-project/opensearch-plugins/issues/238). I initially looked at geospatial since it was one of the plugins that creates an instance of [JobScheduler's LockService](https://github.com/opensearch-project/geospatial/pull/677) and I wanted to demonstrate how to use the instance of LockService provided by job scheduler instead of creating a separate instance. Eventually, I want to remove the [public constructor](https://github.com/opensearch-project/job-scheduler/blob/main/spi/src/main/java/org/opensearch/jobscheduler/spi/utils/LockService.java#L57-L60) for LockService to enforce that plugins use the instance provided by job scheduler.

The reason for this is a new model for access to system indices instead of the current model of wrapping with `ThreadContext.stashContext`.

In the current model there are not authz checks that are run in that block so a plugin can perform any action w/o restriction.

In the new model (utilizing `pluginSubject.runAs`) it provides information at runtime about which plugin is running the action and security uses that information to allow it to access its own system indices, but can forbid other actions.

yizheliu-amazon:
I see. Thank you for the information.

heemin32 (approved):

junqiu-lei (approved):

cwperks:
Fixed the errors due to the breaking change from core.

I added a `public shouldIgnoreMalformed` to `XYShapeFieldMapper` and `XYPointFieldMapper` that can be called in the tests.

PR #766, opened 2025-06-20
Refresh the Ip2Geo cache and retry one more time when we run into an issue

bzhangam:
### Description
Refresh the Ip2Geo cache and retry one more time when we run into an issue.

After we update the datasouce in the primary shard if we fail to send the request to the nodes with replica shard we will not update the cache. Even later the replica is recovered from the primary shard the cache update logic will not be triggered. So and this change to refresh the cache and retry when we run into an issue.

### Related Issues
N/A

### Check List
- [x] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
+++ b/src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
@@ -58,11 +58,20 @@ public class Ip2GeoCachedDao implements IndexingOperationListener {
             .addSettingsUpdateConsumer(Ip2GeoSettings.CACHE_SIZE, setting -> this.geoDataCache.updateMaxSize(setting.longValue()));
     }
 
-    public String getIndexName(final String datasourceName) {
+    private String doGetIndexName(final String datasourceName) {
         return getMetadata().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getIndexName();
     }
 
-    public boolean isExpired(final String datasourceName) {
+    public String getIndexName(final String datasourceName) {
+        String indexName = doGetIndexName(datasourceName);
+        if (indexName == null) {
+            refreshDatasource(datasourceName);
+            indexName = doGetIndexName(datasourceName);
+        }
+        return indexName;
+    }
+
+    private boolean doIsExpired(final String datasourceName) {
         final Instant expirationDate = getMetadata().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getExpirationDate();
         final Instant now = Instant.now();
         final boolean isExpired = expirationDate.isBefore(now);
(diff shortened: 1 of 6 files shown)


heemin32:
I think we should just add retry in these four methods instead of handling each case outside of these methods.
```
  public String getIndexName(final String datasourceName) {
        return getMetadata().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getIndexName();
    }

    public boolean isExpired(final String datasourceName) {
        return getMetadata().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getExpirationDate().isBefore(Instant.now());
    }

    public boolean has(final String datasourceName) {
        return getMetadata().containsKey(datasourceName);
    }

    public DatasourceState getState(final String datasourceName) {
        return getMetadata().getOrDefault(datasourceName, DatasourceMetadata.EMPTY_METADATA).getState();
    }
```

src/main/java/org/opensearch/geospatial/ip2geo/dao/DatasourceDao.java
```diff
@@ -286,6 +286,25 @@ public void getDatasources(final String[] names, final ActionListener<List<Datas
         );
     }
 
+    /**
+     * Get datasources from the primary shard of an index {@code DatasourceExtension.JOB_INDEX_NAME}
+     * @param names the array of datasource names
+     */
+    public List<Datasource> getDatasourcesFromPrimary(final String[] names) {
```
heemin32, line 293: There is a method `public Datasource getDatasource(final String name)` you can use already.

src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
```diff
@@ -122,7 +123,18 @@ private void remove(final String datasourceName) {
...
 
-    private void clearMetadata() {
+    public void refreshDatasource(final String datasourceName){
+        try {
+            List<Datasource> datasourceList =  datasourceDao.getDatasourcesFromPrimary(new String[]{datasourceName});
+            if(datasourceList.getFirst() != null){
+                getMetadata().put(datasourceName, new DatasourceMetadata(datasourceList.getFirst()));
+            }
+        }catch (Exception e){
+            log.error("Fail to refresh the datasource.", e);
```
heemin32, line 133: In this case, we can clean the entire cache?

heemin32:
For 
```
    public Map<String, Object> getGeoData(final String indexName, final String ip) {
        try {
            return geoDataCache.putIfAbsent(indexName, ip, addr -> geoIpDataDao.getGeoIpData(indexName, ip));
        } catch (ExecutionException e) {
            throw new RuntimeException(e);
        }
    }
```
We

src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
bzhangam: yeah

src/main/java/org/opensearch/geospatial/ip2geo/dao/DatasourceDao.java
bzhangam: ok

src/main/java/org/opensearch/geospatial/ip2geo/dao/Ip2GeoCachedDao.java
```diff
@@ -122,6 +171,19 @@ private void remove(final String datasourceName) {
         getMetadata().remove(datasourceName);
     }
 
+    private void refreshDatasource(final String datasourceName) {
+        try {
+            log.info("Refresh datasource.");
+            Datasource datasource = datasourceDao.getDatasource(datasourceName);
+            if (datasource != null) {
+                getMetadata().put(datasourceName, new DatasourceMetadata(datasource));
+            }
```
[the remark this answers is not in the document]
bzhangam: ok

heemin32 (approved):
LGTM. Thanks!

